PR #48, opened 2020-05-19
Notify ViewObservers of existing views

RoanH:
As discussed before the first view is currently created before the plugins have finished loading (unless you pass a file via the command line). Therefore a plugin that implements `ViewObserver` never gets notified of the first view.

This commit fixes that by passing all existing views to a new `ViewObserver` as soon as it is registered. It's also worth noting that a similar code path already exists for plugins of the `newPresentationInterface` type.

I've done some testing and this fix does address the issue I was running into. I ran a `grep` on the code base but couldn't find any existing plugins of the `ViewObserver` to test for possible regressions.


RoanH:
Small note that the Travis CI build fails because the encrypted CI environment variables are removed for PRs (for security reasons).

PR #50, opened 2020-05-24
Add support for CMYK Tiff files

LukeSerne:
This PR extends the plugin that loads TIFF files to also work with CMYK TIFF images. Additionally, it includes a new `Operations` class, `OperationsCMYK`, that handles the 4 channels.


libs/tiled-bitmap/src/cmyklayeroperations.cc
```diff
@@ -0,0 +1,331 @@
...
+int OperationsCMYK::getBpp()
+{
+  // CMYK has 4 channels -> 4 samples per pixel.
+  return this->bps * 4;
+}
+
+// From https://www.pagetable.com/?p=23#comment-1140
+inline uint16_t DivideBy255(uint16_t value)
+{
+  return static_cast<uint16_t>((value + 1 + (value >> 8)) >> 8);
```
kees-jan, line 60: Premature optimization?

inc/scroom/layeroperations.hh
```diff
@@ -142,3 +142,18 @@ public:
   virtual void reduce(Tile::Ptr target, const ConstTile::Ptr source, int x, int y);
 };
 
+class OperationsCMYK : public CommonOperations
```
kees-jan, line 145: - I recommend you make separate classes for the 1, 2, 4 and 8 bps cases. They share hardly any code
- I also recommend you choose your types such that you need far fewer casts. The C++ language designers made casts look ugly (compared to C) to encourage you to avoid them were possible
- Take a look at the [PixelIterator](https://kees-jan.github.io/scroom/0.10/classScroom_1_1Bitmap_1_1PixelIterator.html) class. When used as a SampleIterator, it should save you a fair amount of code
- When reducing a 1, 2 or 4 bps layer, store the result in a 8bps layer. That'll look better and save you from some rounding errors

inc/scroom/layeroperations.hh
```diff
@@ -142,3 +142,18 @@ public:
...
+  static Ptr create(int bps);
+  OperationsCMYK(uint16_t bps);
+  virtual ~OperationsCMYK()
+  {}
+
+  virtual int getBpp();
+  virtual Scroom::Utils::Stuff cache(const ConstTile::Ptr tile);
+  virtual void reduce(Tile::Ptr target, const ConstTile::Ptr source, int x, int y);
+private:
+  uint16_t bps;
```
kees-jan, line 157: Why is this a `uint16_t`? Your default type should be (`unsigned`) `int`. Anything else should have a good reason. To me, this code suggests you consider supporting `bps` values up to 64000 or something?

libs/tiled-bitmap/src/cmyklayeroperations.cc
```diff
@@ -0,0 +1,331 @@
...
+            sum_m += row[current    ] & 15;
+            sum_y += row[current + 1] >> 4;
+            sum_k += row[current + 1] & 15;
+          }
+        }
+
+        targetPtr[0] = static_cast<byte>( sum_c == 15 * 64 ? 0xF0 : (sum_c / 60) << 4)
+                     | static_cast<byte>((sum_m == 15 * 64 ? 0x0F : sum_m / 60));
+        targetPtr[1] = static_cast<byte>( sum_y == 15 * 64 ? 0xF0 : (sum_y / 60) << 4)
+                     | static_cast<byte>((sum_k == 15 * 64 ? 0x0F : sum_k / 60));
```
kees-jan, line 236: - Why are you dividing those sums by 60?
- Why the special case when the sum equals 15*64?

libs/tiled-bitmap/src/cmyklayeroperations.cc
```diff
@@ -0,0 +1,331 @@
...
+            sum_m += (row[current] & 0x30) >> 4;
+            sum_y += (row[current] & 0x0C) >> 2;
+            sum_k +=  row[current] & 0x03;
+          }
+        }
+
+        targetBase[x] = static_cast<byte>(((sum_c == 192 ? 191 : sum_c) / 48) << 6)
+                     | static_cast<byte>(((sum_m == 192 ? 191 : sum_m) / 48) << 4)
+                     | static_cast<byte>(((sum_y == 192 ? 191 : sum_y) / 48) << 2)
+                     | static_cast<byte>(((sum_k == 192 ? 191 : sum_k) / 48)     );
```
kees-jan, line 274: Same questions as above

libs/tiled-bitmap/src/cmyklayeroperations.cc
```diff
@@ -0,0 +1,331 @@
...
+        int sum_c = 0;
+        int sum_m = 0;
+        int sum_y = 0;
+        int sum_k = 0;
+        for (const byte* row = base; row < end; row += sourceStride)
+        {
+          for (size_t current = 0; current < 8; current++)
+          {
+            if ((current & 1) == 0) {
+              sum_c +=  row[current/2]         >> 7;
```
kees-jan, line 301: Yup... You definitely need a PixelIterator

libs/tiled-bitmap/src/cmyklayeroperations.cc
```diff
@@ -0,0 +1,331 @@
...
+#include <sstream>
+
+#include <boost/shared_ptr.hpp>
+#include <boost/utility.hpp>
+
+#include <scroom/bitmap-helpers.hh>
+#include <scroom/cairo-helpers.hh>
+#include <scroom/unused.hh>
+
+#include <boost/math/special_functions/round.hpp>
```
kees-jan, line 25: Where's this one used?

libs/tiled-bitmap/src/cmyklayeroperations.cc
LukeSerne: Moving the division by 255 to a specialised function causes a performance increase of a little over 3% (3.35s -> 3.24s) when loading a 14391 * 6232 8bps TIFF image. Recent improvements have decreased this number though. In a previous version, the performance increase due to this function was almost 15%. As such, I wouldn't call this optimisation premature.

However, whether the 3% load time decrease is worth the reduced readibility is a less objective matter. In my opinion, the readability of the code doesn't suffer enough to change this. If you disagree and think this should not be in a separate function, I'm happy to change it.

inc/scroom/layeroperations.hh
LukeSerne: - I noticed that the RGB Operation classes are all split based on the number of bits per sample. I think they have quite a bit of code duplication, with just a few constants being different. With the "Don't Repeat Yourself" creed in mind, I figured it'd be better to just have one class that handles all different bitdepths. At that time, I noticed the PixelIterator class, and I expected it to be able to abstract those differences away.
- I agree that type casts shouldn't be littered throughout the code. Sometimes however, adding casts avoids warning and overflows. Sometimes, you simply need data types smaller than the default `int`. I think this is one of those cases. Please let me know if you have a good way to save specific casts.
- Unfortunately, I can't get the PixelIterator to work for CMYK. Sadly, not all existing Operations (sub)classes use the PixelIterator and there is not really documentation on how to use it. Prompted by this comment, I tried it again, but to no avail. I could not find any reference to "SampleIterator" in the code / documentation. I guess you meant that I should create a PixelIterator whose "Base" type is a type the individual samples precisely fit in (which would only work for 8bps, unfortunately). However, the number of pixels per byte is now less than 1, which means that it is rounded to 0 because of integer division. This causes various problems, most importantly a division by 0 in the constructor.
- I agree that it helps with rounding (and thus makes it look better). However, I could not figure out how to make sure the Tile created by the reduce function has a different bitdepth from the source Tile.

inc/scroom/layeroperations.hh
LukeSerne: To me, using an `unsigned int` in this case suggests that we're planning to support up to 2147483647 bits per sample. Additionally, the TIFF specification defines the number of bits per sample to be a 16-bit unsigned integer. As such, I think it's only natural to use a `uint16_t`.

libs/tiled-bitmap/src/cmyklayeroperations.cc
LukeSerne: Dividing the sum by 64 (as many of the existing Operation classes do) doesn't quite divide the sum into even groups. Let me give an example for 2bps. Note that the maximum value for the sum is 3 * 64 = 192. For 4bps, the maximum value is 15 * 64.

Division by 64:
sum    0 ... 63 64 ... 127 128 ... 191 192
result 0 ... 0   1 ... 1     2 ... 2   3

"Fairer":
sum    0 ... 47 48 ... 95 96 ... 143 144 ... 191 192
result 0 ... 0   1 ... 1   2 ... 2     3 ... 3   [4 -> 3]

This alternative implementation has made the groups of values that map to a specific value equally big (except for the largest result value). The difference between these alternatives is smaller as the number of bits per sample grows. In this 2bps case, the constants would be 48 and 192. So the constants 60 and 15 * 64 (for 4bps) come from this alternative approach.

libs/tiled-bitmap/src/cmyklayeroperations.cc
LukeSerne: Same answer as above

libs/tiled-bitmap/src/cmyklayeroperations.cc
LukeSerne: See the third bullet point in the answer to your second comment on this file.

libs/tiled-bitmap/src/cmyklayeroperations.cc
LukeSerne: See the third bullet point in the answer to your second comment on this file.

libs/tiled-bitmap/src/cmyklayeroperations.cc
LukeSerne: Good catch. I copied the includes from `layeroperations.cc` without properly checking that all of them were still required. I pushed a new commit that removed all include statements that are not needed.

libs/tiled-bitmap/src/cmyklayeroperations.cc
kees-jan: For only 3%, I would not choose to use constructs like this one.

inc/scroom/layeroperations.hh
kees-jan: _Regarding splitting this class_
As it is, these methods are fairly large. Near as I can tell, there would be hardly any shared code if you split them. If you decide to go the other way and abstract the differences away, that’s also fine. It is just that, as it is, it is not very maintainable, and I can’t imagine it would score well on the metrics laid out by the university.

_Regarding PixelIterator_
If there’s anything in the codebase that you don’t understand or can’t get to work, feel free to ask, I’m happy to help. When you do ask questions, please make them as specific as possible, detailing what you tried, what you expected to happen, and what actually happen. “Unfortunately there is no documentation” (technically not even a question :wink:) is a bit to broad to tackle in limited time

I’ve noticed that there aren’t yet any tests for the PixelIterator class. If you were to contribute some tests that are failing, even though in your opinion they shouldn’t be, I’d be happy to look at them.

_Regarding casts_
The more different integer types you use, the more you will need to cast them. My recommendation would be to stick to (unsigned) integer, and only cast them when storing in the target memory location

The default should be that there are no casts at all. That means that it is not my task to suggest specific improvements, but yours to defend any casts that are currently in the code.

_Regarding rounding and layers with different bpp values_
Just add a second LayerOperations object with a different bpp value to your LayerSpec. Then the subsequent layers will have that bpp value. See the 1bpp grayscale implementation for an example

inc/scroom/layeroperations.hh
kees-jan: I think I would be in favor of that

libs/tiled-bitmap/src/cmyklayeroperations.cc
kees-jan: Ouch
1. That approach could do with a comment. Or better yet, a properly named function that implements this
2. Still confused. 144/60 is 2.4, which becomes 2, not 3 as you suggest in the table
3. To arrive at a sum of 144, you need 48 2's and only 16 3's. That should really become a 2 not a 3.

Summarizing: I'm not opposed to rounding, but I'd like to see a more conventional algorithm. 32 2's and 32 3's surely should become a 3, in the interest of "fairness", but everything lighter than that should be a 2 or less.

LukeSerne: My example was for 2bps, whereas this code is for 4bps, so the constants would be slightly different. Then the math does work :wink:.

Regardless, this construct is no longer needed when higher layers always use 8bps (as they do since commit 27e92a12244c6fb6408e07e0f78329e48c153ada. As such, I think we can resolve this conversation.

inc/scroom/layeroperations.hh
LukeSerne: _Regarding splitting `OperationsCMYK`_
I've split the class up in 4 classes in commit 27e92a12244c6fb6408e07e0f78329e48c153ada. I think that resolves that bullet point.

_Regarding PixelIterator_
I think making separate classes for each of the different bitdepths removes the need for a `PixelIterator`. I'm still unsure what you meant by `SampleIterator`.

_Regarding casts_
By moving the various bitdepths to different classes, most casts have disappeared. As such, I also think that this bullet point is resolved.

kees-jan: I think this code still has unneeded casts, and could still benefit from a PixelIterator.

But I'm giving up, for now :wink:

PR #52, opened 2020-05-28
Allow plugins to load presentations

RoanH:
The aim of this change was to enable plugins to load presentations on their own. The main use case for this being a plugin that does something with a set of presentations, a possible alternative use case a plugin that changes the presentation on display in a view. 

Currently compiling against another plugin and directly calling that plugin to parse a file to open a presentation is the best alternative. However, this is a solution that fails if a plugin is not present and limited to plugins that are known to exist. The solution in this PR allows a plugin to ask Scroom to open any file and Scroom will figure out which plugin type is capable of opening the file.

When trying to add a code path for this the `ScroomPluginInterface` seemed to me as the best place to add this as it is the only thing shared by all plugin types that also has the ability interact directly with Scroom. I'm not sure however, if a better alternative may have been available. This would also be the first addition to the `ScroomPluginInterface` that is not strictly for registering a new plugin.

Anyway, let me know what you think @kees-jan (did that trigger an email notification?).

### DIFF
--- a/gui/src/callbacks.cc
+++ b/gui/src/callbacks.cc
@@ -603,3 +603,7 @@ void on_new_viewobserver(ViewObserver::Ptr v)
   }
 }
 
+PresentationInterface::Ptr on_load_presentation(const std::string& filename)
+{
+  return loadPresentation(filename);
+}
--- a/gui/src/callbacks.hh
+++ b/gui/src/callbacks.hh
@@ -78,3 +78,5 @@ void on_new_presentationobserver(PresentationObserver::Ptr po);
 void on_new_viewobserver(ViewObserver::Ptr v);
 
 void on_presentation_possibly_destroyed();
+
+PresentationInterface::Ptr on_load_presentation(const std::string& filename);
--- a/gui/src/pluginmanager.cc
+++ b/gui/src/pluginmanager.cc
@@ -247,6 +247,11 @@ void PluginManager::registerPresentationObserver(const std::string& identifier, 
   presentationObservers[observer] = identifier;
 }
 
+PresentationInterface::Ptr PluginManager::loadPresentation(const std::string& filename)
+{
+  return on_load_presentation(filename);
+}
+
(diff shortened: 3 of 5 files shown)


kees-jan:
Thank you for that pull request. Had to think long and hard about it, and reconsider the design of scroom, which is always an interesting exercise. (And yes, the mentioning worked - still have to figure out why I don't get an email otherwise, though)

As you already hinted `ScroomPluginInterface` is not the correct location for this method call. Reason is that in order to actually call that method, the plugin would need to store a reference to the `ScroomPluginInterface`, such that it can be used in a later call to `OpenPresentationInterface::open()`. However, if you do that, you'd create a cycle of shared pointers (`PluginManager` to the plugin, and the plugin to the `PluginManager`, via `ScroomPluginInterface`, which would result in a memory leak, possibly even problems closing presentations or scroom.

The solution to this is to use the `ScroomInterface` instead, which already has a method `loadPresentation()`, which does exactly what you want. You now have two options:
- Extend `OpenPresentationInterface::open()`, such that the plugin receives a `ScroomInterface::Ptr`, with which it can open other presentations
- Have your plugin implement `OpenInterface`, instead of `OpenPresentationInterface`. `OpenInterface::open()` already as a `ScroomInterface::Ptr` argument included exactly for the purpose you need.

Having said all that, I don't quite see how doing this is going to help you implement the plugins you want. Assuming that you're working on the sep plugin, after loading the presentations, you'll have no way to combine the colors in the way Canon wants, unless you can get at the bitmap data, which is not possible to do through the `PresentationInterface`. And even if you were to fix that by casting to another interface (which I would not be in favor of), then you'd still have the practical problem that the tile you need might not be in memory. In fact, there is no check that the presentation is in fact CMYK, has the correct bit depth, or even actually is a bitmap.

If you like, we can have a short design discussion after the demo today?

RoanH:
I'll close this for now since the issues are clear and this should not get merged as is.

PR #57, opened 2020-05-29
WIP: Suppress most library warnings on Windows, add icon

ethoxx:
Suppresses most of the library warnings on Windows by including the libraries with `-isystem` instead of `-I`. A few warnings coming from boost remain, these don't seem to be that easily fixable though.
Furthermore, the bald eagle has been added as the Scroom icon on Windows.


configure.ac
```diff
@@ -113,14 +113,14 @@ DX_PS_FEATURE(OFF)
...
+PKG_CHECK_MODULES([UTILS], [gtk+-2.0 >= 2.0.0], [UTILS_CFLAGS=$(echo "$UTILS_CFLAGS" | sed s/-I/$ISYSTEM_FLAG\ /g)])
+PKG_CHECK_MODULES([TILEDBITMAP], [gtk+-2.0 >= 2.0.0, cairo], [TILEDBITMAP_CFLAGS=$(echo "$TILEDBITMAP_CFLAGS" | sed s/-I/$ISYSTEM_FLAG\ /g)])
+PKG_CHECK_MODULES([SCROOMLIB], [gtk+-2.0 >= 2.0.0, cairo], [SCROOMLIB_CFLAGS=$(echo "$SCROOMLIB_CFLAGS" | sed s/-I/$ISYSTEM_FLAG\ /g)])
+PKG_CHECK_MODULES([EXAMPLEPLUGIN], [gtk+-2.0 >= 2.0.0], [EXAMPLEPLUGIN_CFLAGS=$(echo "$EXAMPLEPLUGIN_CFLAGS" | sed s/-I/$ISYSTEM_FLAG\ /g)])
+PKG_CHECK_MODULES([TRANSPARENTOVERLAYPLUGIN], [gtk+-2.0 >= 2.0.0], [TRANSPARENTOVERLAYPLUGIN_CFLAGS=$(echo "$TRANSPARENTOVERLAYPLUGIN_CFLAGS" | sed s/-I/$ISYSTEM_FLAG\ /g)])
+PKG_CHECK_MODULES([COLORMAPPLUGIN], [gtk+-2.0 >= 2.0.0, cairo], [COLORMAPPLUGIN_CFLAGS=$(echo "$COLORMAPPLUGIN_CFLAGS" | sed s/-I/$ISYSTEM_FLAG\ /g)])
+PKG_CHECK_MODULES([TIFFPLUGIN], [gtk+-2.0 >= 2.0.0, cairo], [TIFFPLUGIN_CFLAGS=$(echo "$TIFFPLUGIN_CFLAGS" | sed s/-I/$ISYSTEM_FLAG\ /g)])
+PKG_CHECK_MODULES([GUI],        [gtk+-2.0 >= 2.0.0, libglade-2.0, gthread-2.0, gmodule-2.0, cairo], [GUI_CFLAGS=$(echo "$GUI_CFLAGS" | sed s/-I/$ISYSTEM_FLAG\ /g)])
```
kees-jan, line 123: A variant of this has been attempted before, during the windows port. The end result is that all system include directories are prefixed by `-isystem`, which is undesirable in C++. When using templates incorrectly, that will trigger a warning in the included headerfile, and you'll want to see those. Hence, directories containing C++ includes cannot be included with `-isystem`

If you were to restrict `-isystem` usage to the directories containing gtk/gdk/cairo headers, that would be fine. Alternatively, see [this pull request](https://gitlab.com/sep-group-2/scroom/-/merge_requests/29) for my approach on reducing warnings.

ethoxx: That's a pretty good point. As the icon compilation is still not working here, I will just close the PR and make a new one to prevent all the unnecessary commits from being merged.

PR #58, opened 2020-05-30
Add icon on Windows

ethoxx:


ethoxx:
This branch builds successfully on both my local machine and Gitlab CI. I have no idea why it doesn't work on Travis...

Makefile.am
```diff
@@ -571,6 +571,10 @@ gui_src_scroom_SOURCES = \
 	gui/src/progressbarmanager.hh \
 	gui/src/workinterface.hh
 
+if HAVE_WINDRES
+gui_src_scroom_SOURCES += build_scripts/windows/icon.rc
+endif
```
kees-jan, line 576: Theoretically, it should be possible to do "make dist" on linux, and then build the resulting archive on windows. That means that the icon.rc should be in the tgz, even on Linux. That's currently not happening

I recommend adding icon.rc to gui_extra_dist, or something, either unconditionally, or in case we don't have windres. That should take care of that.

Note that doing "make dist" on linux and then building on windows doesn't currently work. The generated configure script doesn't work as expected. Still, no reason to make thing worse ;-)

kees-jan:
Oh, I apologize for not looking at this pull request sooner. I was distracted by questions about SampleIterators.

Makefile.am
ethoxx: Alright, I added `icon.rc` to gui_extra_dist

PR #59, opened 2020-05-31
View Additions

RoanH:
Hello @kees-jan,

As indicated before here's the first of several PRs related to the pipette tool.

To keep all the changes a bit more managable we've decided to make 3 separate self-contained pull requests.
- This pull request is for the additional `View` functionality and the extraction of the `Measure` tool.
- A pull request for the functionality that allows a plugin to get average color data from a presentation.
- A pull request dependent on the other two pull requests with the actual pipette tool. Since the last meeting the exact implementation details of the pipette tool are a bit uncertain so we'll first confirm the exact details with Canon.

In order to streamline the merging process around this pull request I'll try to justify all the changes and concerns.

### Changes
1. Addition of `PostRenderer`: Can be registered via the `ViewInterface`, currently only `View` implements all the functions required for it. A `PostRenderer` is executed after the `Presentation` referenced by a `View` finishes drawing. A `PostRenderer` is passed the `cairo` instance so it can draw something.
2. Addition of a `setPanning` and `unsetPanning` function: The current specification of the pipette tool in the URD requires us to bind the tool to the left mouse button. Since this would otherwise conflict with panning we need the option to disable panning while the pipette tool is active. It might be that this is no longer an issue after discussing the details of the pipette tool with Canon.
3. Addition of `SelectionListener`: Can be registered via the `ViewInterface`, currently only `View` implements all the functions required for it. The `SelectionListener` is registered for some `MouseButton` and is called on three events: When the user presses the mouse button, when the selection updates due to dragging, and when the mouse button is released again.
4. The measure tool was turned into its own plugin and uses the newly implemented functionality.
5. The `Measurement` struct was removed from `view.hh` and reintroduced as `Selection` in `viewinterface.hh`.
6. The `presentationPointToWindowPoint` function in `View` was exposed via the `ViewInterface` so `PostRenderer` implementations can use it to draw things based on `Selections` received by `SelectionListeners`.
7. The `setStatusMessage` function in `View` was exposed via the `ViewInterface` so plugins can set a status message.
8. Addition of a `MouseButton` enum, used when registering a `SelectionListener`.
9. Addition of a `getCurrentPresentation` function to the `ViewInterface`. Required by the pipette tool to get the presentation and then the color data from the presentation. See consideration 7 though.

### Considerations
1. Additions to the `ViewInterface` have been given empty bodies or default implementations instead of making them pure virtual functions. This is because we made a promise (in the URD) to not make changes that are not backwards compatible.
2. `Measure` has been renamed to `Selection` and reused for the `SelectionListener`. We tried using a `Scroom::Utils::Rectangle` instead. However, this meant that we lost information about the start and end position. This meant that for dragging we would not know which point to update. This also meant that plugins that implement `SelectionListener` would no longer know about the start and end position (which for `Measure` is used to draw the diagonal). It would be possible to work around all of this, but this would make both plugins and `View` more complex.
3. In light of my other PR, `View` currently holds a map of shared pointers to registered `SelectionListeners` and `PostRenderers`, however most `SelectionListeners` and `PostRenderers` will need a reference to the `ViewInterface`, this means they want to store a shared pointer to the `ViewInterface`. However, I'm now worried that this presents a similar shared pointer cycle as was present in my previous PR. Within our team we weren't able to come up with a proper answer to this concern though. If this is in fact a cycle then a possible fix would be to just pass a shared pointer to the `ViewInterface` with all the events. Then there would be no need for a plugin to store a shared pointer to the `ViewInterface`. Let me know if this is something I should change.
4. Panning was not extracted to be its own plugin. I felt like a lot of the `View` internals would have to be exposed to do this and I wasn't comfortable doing so without any prior discussion.
5. On line 697~712 in `view.cc` I need to mask the event state with all the button masks. Although the current implementation works fine, I feel like there should be a cleaner way to do this.
6. The `MouseButton` enum currently only covers 3 buttons, there is no documentation for other buttons, but perhaps it would be better to add them anyway. Mostly these buttons will be macro buttons on newer mice.
7. The `getCurrentPresentation` function returns a `Scroom::Utils::Stuff` because it is not possible to include the required header for `PresentationInterface::Ptr` in `viewinterface.hh` as this would be a cyclic include. This solution is not ideal, but we also do not have any better ideas.

We tried to match the general design style closely to existing things, but we might have overlooked something. If you have any further questions I'll try my best to answer them.

Best regards,
Roan


plugins/measure/src/measure.cc
```diff
@@ -0,0 +1,132 @@
...
+////////////////////////////////////////////////////////////////////////
+
+void MeasureHandler::render(cairo_t* cr)
+{
+  if(selection)
+  {
+    GdkPoint start = view->presentationPointToWindowPoint(selection->start);
+    GdkPoint end = view->presentationPointToWindowPoint(selection->end);
+    cairo_set_line_width(cr, 1);
+    cairo_stroke(cr);
```
kees-jan, line 122: You are stroking a previously set-up path? Why?

plugins/measure/src/measure.hh
```diff
@@ -0,0 +1,79 @@
...
+  MeasureHandler();
+
+public:
+  typedef boost::shared_ptr<MeasureHandler> Ptr;
+
+private:
+  Selection* selection;
+
+public:
+  ViewInterface::Ptr view;
```
kees-jan, line 26: You cannot store a `ViewInterface` pointer in anything that is kept alive by a `View`.

I would prefer that you pass in a `ViewInterface::Ptr` on methods where it is needed. Alternatively, store a `weak_ptr`

kees-jan:
> Hello @kees-jan,
> 
> As indicated before here's the first of several PRs related to the pipette tool.
Thanks! Love smaller pull requests (and this one is already quite big :wink: )

My apologies for not starting on this PR earlier. I was distracted by some questions about `SampleIterator`s.

> ### Changes
> 2. Addition of a `setPanning` and `unsetPanning` function: The current specification of the pipette tool in the URD requires us to bind the tool to the left mouse button. Since this would otherwise conflict with panning we need the option to disable panning while the pipette tool is active. It might be that this is no longer an issue after discussing the details of the pipette tool with Canon.

Note that Pipette tool needs only be bound to the left mouse button in "Adobe" profile. In the default profile, we're free to do what is convenient.

> 3. Addition of `SelectionListener`: Can be registered via the `ViewInterface`, currently only `View` implements all the functions required for it. The `SelectionListener` is registered for some `MouseButton` and is called on three events: When the user presses the mouse button, when the selection updates due to dragging, and when the mouse button is released again.

I think registering for  selection events for a given mouse button is a bad idea. It makes it harder to distinguish "adobe" and "default" profiles later.

There are several ways you can go
* Pipette and measure tool only control the content that is shown on selection event, but not the selection process itself (as proposed elsewhere)
* When Pipette tool is selected, it should receive selection events from the right button only (consistent with current behavior - to be changed when adobe profile is introduced), and the Measure tool is turned off
* When Pipette tool is selected, it should receive selection events from both mouse buttons (Weird choice that might somehow appeal to Canon)

> 4. The measure tool was turned into its own plugin and uses the newly implemented functionality.

Shouldn't there be a button for the measure tool? Otherwise you'll get a conflict later as to who receives the selection event and/or who gets to update the statusbar

> 6. The `presentationPointToWindowPoint` function in `View` was exposed via the `ViewInterface` so `PostRenderer` implementations can use it to draw things based on `Selections` received by `SelectionListeners`.

Not sure that was wise. See my suggestions elsewhere for alternatives

> 7. The `setStatusMessage` function in `View` was exposed via the `ViewInterface` so plugins can set a status message.

That'll work, but only if measure and pippete tool are mutually exclusive. Otherwise you have a conflict

> 8. Addition of a `MouseButton` enum, used when registering a `SelectionListener`.

Considering that I'm not a fan of having the plugin be in control of which button is involved, I'd say this is not necessary :wink: 

> 9. Addition of a `getCurrentPresentation` function to the `ViewInterface`. Required by the pipette tool to get the presentation and then the color data from the presentation. See consideration 7 though.

Storing references to presentations is fine. It is the references to `View`s you should worry about.

> ### Considerations
> 1. Additions to the `ViewInterface` have been given empty bodies or default implementations instead of making them pure virtual functions. This is because we made a promise (in the URD) to not make changes that are not backwards compatible.

Considering that the only other `ViewInterface` implementation is in the transparent overlay, you might as well move them there. That way, if you were to insist on keeping `presentationPointToWindowPoint()` public, you could do a proper implementation there.

> 2. `Measure` has been renamed to `Selection` and reused for the `SelectionListener`. We tried using a `Scroom::Utils::Rectangle` instead. However, this meant that we lost information about the start and end position. This meant that for dragging we would not know which point to update. This also meant that plugins that implement `SelectionListener` would no longer know about the start and end position (which for `Measure` is used to draw the diagonal). It would be possible to work around all of this, but this would make both plugins and `View` more complex.

Not sure though why you chose to draw both the rectangle and the line.

I feel that for a Measure tool, a straight line would be most natural, though I don't object to it being a rectangle. For a Pipette tool, a rectangle makes most sense, so for ease of implementation you might choose to have them both a rectangle.

However, ease of implementation is not an issue, since both will have their own `PostRenderer`. And I never saw any proposal or reasoning for drawing both the line and the rectangle.

I'm not necessarily opposed, it is just unconventional. And doing unconventional stuff needs (in my opinion) sound reasoning, and I haven't seen it yet :wink: 

> 3. In light of my other PR, `View` currently holds a map of shared pointers to registered `SelectionListeners` and `PostRenderers`, however most `SelectionListeners` and `PostRenderers` will need a reference to the `ViewInterface`, this means they want to store a shared pointer to the `ViewInterface`. However, I'm now worried that this presents a similar shared pointer cycle as was present in my previous PR.

It does. See my feedback on you `MeasureHandler` implementation

> 4. Panning was not extracted to be its own plugin. I felt like a lot of the `View` internals would have to be exposed to do this and I wasn't comfortable doing so without any prior discussion.

Not sure. If, on `ViewInterface`, there were `SetCenterCoordinate()` and `GetCenterCoordinate()` methods, then a "PanningPlugin" would need to subscribe to selection events (left mouse button), and call those two methods as appropriate.

Functionality-wise, this is not different from what the two textboxes in the toolbar are already doing. Updating scrollbars and rulers and stuff should remain the responsibility of the `View`

> 5. On line 697~712 in `view.cc` I need to mask the event state with all the button masks. Although the current implementation works fine, I feel like there should be a cleaner way to do this.

I share your worry, but given the other feedback, this is not a priority. In fact, it may change "automatically" when you pick up some of the other items.

> 6. The `MouseButton` enum currently only covers 3 buttons, there is no documentation for other buttons, but perhaps it would be better to add them anyway. Mostly these buttons will be macro buttons on newer mice.

As said elsewhere, plugins should (in my opinion) not be concerned with mouse buttons

> 7. The `getCurrentPresentation` function returns a `Scroom::Utils::Stuff` because it is not possible to include the required header for `PresentationInterface::Ptr` in `viewinterface.hh` as this would be a cyclic include. This solution is not ideal, but we also do not have any better ideas.

The forward declaration (suggested there) should work.

> We tried to match the general design style closely to existing things, but we might have overlooked something. If you have any further questions I'll try my best to answer them.

Overall, this looks pretty good. My compliments :+1:

kees-jan:
I pushed a change to `ENVIRONMENT.in`, to make sure that the new plugin works in development mode.

I apologize, It was an accident (didn't even know I could do that). The change is good, though, and warmly recommended :wink:

RoanH:
Thanks for the feedback.

Being able to push to pull requests is a feature you have to enable as the one making the pull request. Since I usually don't mind I have it enabled. However it seems like your commit got force pushed away by the mirroring software that synchronizes our GitLab repository with our GitHub one... I managed to dig up the commit though thanks to GitHub and git so I pushed it back.

.gitignore
```diff
@@ -55,4 +55,5 @@ Makefile.in
 /doc/
 /.google
 test_results.xml
-/.vscode
\ No newline at end of file
+/scroom-build
```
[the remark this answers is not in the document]
RoanH: It's actually the folder I compile scroom itself in. In hindsigh the location is terrible, but I was just exploring the codebase at the time and it stuck around. Not planning on building anything else in the scroom source tree though, so I guess I could've just called it `build`. I can remove it if you'd like.

gui/src/view.hh
```diff
@@ -145,18 +133,22 @@ public:
   virtual void removeSideWidget(GtkWidget* w);
   virtual void addToToolbar(GtkToolItem* ti);
   virtual void removeFromToolbar(GtkToolItem* ti);
+  virtual void setPanning();
+  virtual void unsetPanning();
+  virtual void registerSelectionListener(SelectionListener::Ptr listener, MouseButton button);
+  virtual void registerPostRenderer(PostRenderer::Ptr renderer);
+  virtual void setStatusMessage(const std::string& message);
+  virtual GdkPoint presentationPointToWindowPoint(GdkPoint presentationpoint);
```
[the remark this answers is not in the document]
RoanH: That's a pretty good idea, I'll look into this.

inc/scroom/viewinterface.hh
```diff
@@ -7,14 +7,124 @@
...
 #include <boost/shared_ptr.hpp>
 #include <boost/weak_ptr.hpp>
 
 #include <scroom/progressinterface.hh>
+#include <scroom/stuff.hh>
+
+/**
+ * Structure that represents a selection made
+ * by the user. This structure was previously
+ * called 'Measurement'.
```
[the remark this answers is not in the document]
RoanH: Interesting how I missed that before PRing... that should be changed yeah.

inc/scroom/viewinterface.hh
```diff
@@ -24,7 +134,7 @@
  * presentation, you typically want to influence whatever is being
  * shown. This interface allows you to do so.
  *
- * @see PresentationInterface, Viewable
+ * @see PresentationInterface, SelectionListener, PostRenderer, Viewable
```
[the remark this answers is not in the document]
RoanH: I don't know how this order came about, but I can `SelectionListener` and `PostRenderer` to be last in the list.

inc/scroom/viewinterface.hh
```diff
@@ -88,5 +198,78 @@ public:
...
+  virtual void setStatusMessage(const std::string&)
+  {
+  }
+
+  /**
+   * Converts a point on screen to a presentation pixel.
+   */
+  virtual GdkPoint presentationPointToWindowPoint(GdkPoint)
+  {
+    return {0, 0};
```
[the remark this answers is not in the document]
RoanH: I agree that this is a bad idea, but I'm not allowed to change this. I like the prosed other approaches though so I'll probably remove the entire function later.

plugins/measure/src/measure.cc
```diff
@@ -0,0 +1,132 @@
...
+#include "measure.hh"
+
+////////////////////////////////////////////////////////////////////////
+// Measure
+////////////////////////////////////////////////////////////////////////
+
+Measure::Measure()
+{
+}
+Measure::~Measure()
```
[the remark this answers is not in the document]
RoanH: Followed the tiff/example plugin on this one, switched it over to how transparent-overlay handles it now.

plugins/measure/src/measure.cc
RoanH: Forgot to remove it together with the crosses it seems, removed it now.

plugins/measure/src/measure.hh
RoanH: Figured that might have been the case, I'll add them to the methods since implementations will likely need the view for something either way.

RoanH:
> > 1. Addition of a `setPanning` and `unsetPanning` function: The current specification of the pipette tool in the URD requires us to bind the tool to the left mouse button. Since this would otherwise conflict with panning we need the option to disable panning while the pipette tool is active. It might be that this is no longer an issue after discussing the details of the pipette tool with Canon.
> 
> Note that Pipette tool needs only be bound to the left mouse button in "Adobe" profile. In the default profile, we're free to do what is convenient.

Looks like I misremembered this, the current URD does not have any requirements on the mouse button for pipette (not for the default profile and also not for the adobe profile).

> > 1. Addition of `SelectionListener`: Can be registered via the `ViewInterface`, currently only `View` implements all the functions required for it. The `SelectionListener` is registered for some `MouseButton` and is called on three events: When the user presses the mouse button, when the selection updates due to dragging, and when the mouse button is released again.
> 
> I think registering for selection events for a given mouse button is a bad idea. It makes it harder to distinguish "adobe" and "default" profiles later.
> 
> There are several ways you can go
> 
>     * Pipette and measure tool only control the content that is shown on selection event, but not the selection process itself (as proposed elsewhere)
> 
>     * When Pipette tool is selected, it should receive selection events from the right button only (consistent with current behavior - to be changed when adobe profile is introduced), and the Measure tool is turned off
> 
>     * When Pipette tool is selected, it should receive selection events from both mouse buttons (Weird choice that might somehow appeal to Canon)

Given that binding something to the left mouse button is not a requirement my preferred method would be to bind all the plugins to the same button and make the plugins mutually exclusive. For the time being that could be the right mouse button, if different profiles are implemented later something should be introduced so that the trigger can be changed. The only issue is making the plugins mutually exclusive, since plugins don't know about other plugins this would have to be handled by the view. It might make more sense in this case to have the enable/disable buttons added by the View. Otherwise I see no easy way to make them mutually exclusive.

> > 1. The measure tool was turned into its own plugin and uses the newly implemented functionality.
> 
> Shouldn't there be a button for the measure tool? Otherwise you'll get a conflict later as to who receives the selection event and/or who gets to update the statusbar

Fair, didn't add one to stick closer to the existing implementation but a toggle should probably be added. Also related to the previous remark.

> > 1. The `presentationPointToWindowPoint` function in `View` was exposed via the `ViewInterface` so `PostRenderer` implementations can use it to draw things based on `Selections` received by `SelectionListeners`.
> 
> Not sure that was wise. See my suggestions elsewhere for alternatives

The alternatives should address this yeah.

> > 1. The `setStatusMessage` function in `View` was exposed via the `ViewInterface` so plugins can set a status message.
> 
> That'll work, but only if measure and pippete tool are mutually exclusive. Otherwise you have a conflict

I'm in favour of making selection plugins mututally exclusive, but this is up for discussion.

> > 1. Addition of a `MouseButton` enum, used when registering a `SelectionListener`.
> 
> Considering that I'm not a fan of having the plugin be in control of which button is involved, I'd say this is not necessary 😉

Should be possible to remove this after some of the other changes are done.

> > 1. Addition of a `getCurrentPresentation` function to the `ViewInterface`. Required by the pipette tool to get the presentation and then the color data from the presentation. See consideration 7 though.
> 
> Storing references to presentations is fine. It is the references to `View`s you should worry about.

Alright.

> > ### Considerations
> > 
> > 1. Additions to the `ViewInterface` have been given empty bodies or default implementations instead of making them pure virtual functions. This is because we made a promise (in the URD) to not make changes that are not backwards compatible.
> 
> Considering that the only other `ViewInterface` implementation is in the transparent overlay, you might as well move them there. That way, if you were to insist on keeping `presentationPointToWindowPoint()` public, you could do a proper implementation there.

I would be in favour of this, but I would also technically be breaking a 'must have' in our URD. If you say its fine however then I would be in favour of changing it.

> > 1. `Measure` has been renamed to `Selection` and reused for the `SelectionListener`. We tried using a `Scroom::Utils::Rectangle` instead. However, this meant that we lost information about the start and end position. This meant that for dragging we would not know which point to update. This also meant that plugins that implement `SelectionListener` would no longer know about the start and end position (which for `Measure` is used to draw the diagonal). It would be possible to work around all of this, but this would make both plugins and `View` more complex.
> 
> Not sure though why you chose to draw both the rectangle and the line.
> I feel that for a Measure tool, a straight line would be most natural, though I don't object to it being a rectangle. For a Pipette tool, a rectangle makes most sense, so for ease of implementation you might choose to have them both a rectangle.
> 
> However, ease of implementation is not an issue, since both will have their own `PostRenderer`. And I never saw any proposal or reasoning for drawing both the line and the rectangle.
> 
> I'm not necessarily opposed, it is just unconventional. And doing unconventional stuff needs (in my opinion) sound reasoning, and I haven't seen it yet 😉

The reason for this is that I prefer your original implementation. However, our URD requirement states "The user can select a rectangular area from the image.". The current implementation is a compromise of both.

> > 1. In light of my other PR, `View` currently holds a map of shared pointers to registered `SelectionListeners` and `PostRenderers`, however most `SelectionListeners` and `PostRenderers` will need a reference to the `ViewInterface`, this means they want to store a shared pointer to the `ViewInterface`. However, I'm now worried that this presents a similar shared pointer cycle as was present in my previous PR.
> 
> It does. See my feedback on you `MeasureHandler` implementation

Will be changed.

> > 1. Panning was not extracted to be its own plugin. I felt like a lot of the `View` internals would have to be exposed to do this and I wasn't comfortable doing so without any prior discussion.
> 
> Not sure. If, on `ViewInterface`, there were `SetCenterCoordinate()` and `GetCenterCoordinate()` methods, then a "PanningPlugin" would need to subscribe to selection events (left mouse button), and call those two methods as appropriate.
> 
> Functionality-wise, this is not different from what the two textboxes in the toolbar are already doing. Updating scrollbars and rulers and stuff should remain the responsibility of the `View`

That's fair, the only requirement would be binding to a specific mouse button which we're removing. So this would also have to use the system we chooce to replace that.

> > 1. On line 697~712 in `view.cc` I need to mask the event state with all the button masks. Although the current implementation works fine, I feel like there should be a cleaner way to do this.
> 
> I share your worry, but given the other feedback, this is not a priority. In fact, it may change "automatically" when you pick up some of the other items.

Assuming that this will resolve itself for now.

> > 1. The `MouseButton` enum currently only covers 3 buttons, there is no documentation for other buttons, but perhaps it would be better to add them anyway. Mostly these buttons will be macro buttons on newer mice.
> 
> As said elsewhere, plugins should (in my opinion) not be concerned with mouse buttons

Might be best to discuss this during the meeting later today.

> > 1. The `getCurrentPresentation` function returns a `Scroom::Utils::Stuff` because it is not possible to include the required header for `PresentationInterface::Ptr` in `viewinterface.hh` as this would be a cyclic include. This solution is not ideal, but we also do not have any better ideas.
> 
> The forward declaration (suggested there) should work.

Luke said he was going to try this so I'm awaiting the result for now.

kees-jan:
> Given that binding something to the left mouse button is not a requirement my preferred method would be to bind all the plugins to the same button and make the plugins mutually exclusive. For the time being that could be the right mouse button, if different profiles are implemented later something should be introduced so that the trigger can be changed. The only issue is making the plugins mutually exclusive, since plugins don't know about other plugins this would have to be handled by the view. It might make more sense in this case to have the enable/disable buttons added by the View. Otherwise I see no easy way to make them mutually exclusive.

How about on the `ViewInterface` we add a `addToolButton()`
* It is up to the View to make sure that at most one toolbutton is enabled
* In the `addToolButton()` call, the presentation has to provide a callback function (to be called by the View), such that it knows whether the tool is currently enabled

> I would be in favour of this, but I would also technically be breaking a 'must have' in our URD. If you say its fine however then I would be in favour of changing it.

The background of the "must have" requirement is that Canon doesn't want you to break any code that may already exist. In this case, that will not be the case. So if the function remains exposed, please make the change :wink: 

> The reason for this is that I prefer your original implementation. However, our URD requirement states "The user can select a rectangular area from the image.". The current implementation is a compromise of both.

But that would apply to Pipette tool only, right?

RoanH:
> 
> 
> > Given that binding something to the left mouse button is not a requirement my preferred method would be to bind all the plugins to the same button and make the plugins mutually exclusive. For the time being that could be the right mouse button, if different profiles are implemented later something should be introduced so that the trigger can be changed. The only issue is making the plugins mutually exclusive, since plugins don't know about other plugins this would have to be handled by the view. It might make more sense in this case to have the enable/disable buttons added by the View. Otherwise I see no easy way to make them mutually exclusive.
> 
> How about on the `ViewInterface` we add a `addToolButton()`
> 
>     * It is up to the View to make sure that at most one toolbutton is enabled
> 
>     * In the `addToolButton()` call, the presentation has to provide a callback function (to be called by the View), such that it knows whether the tool is currently enabled

Sounds good to me, I'll look into this.

> > I would be in favour of this, but I would also technically be breaking a 'must have' in our URD. If you say its fine however then I would be in favour of changing it.
> 
> The background of the "must have" requirement is that Canon doesn't want you to break any code that may already exist. In this case, that will not be the case. So if the function remains exposed, please make the change 😉

It would technically break any 3rd party plugins that implement the `ViewInterface` given that the transparent overlay plugin exists I assumed that this would be a possibility and therefore a breaking change. But since you're fine with it I'll make the change.

> > The reason for this is that I prefer your original implementation. However, our URD requirement states "The user can select a rectangular area from the image.". The current implementation is a compromise of both.
> 
> But that would apply to Pipette tool only, right?

They also requested a selection tool.
afbeelding
We are currently suggesting to them to drop the top-left and bottom-right corner part so that we can use start and end instead. An alternative would be to make it a separate plugin I suppose.

RoanH:
In their latest email Canon indicated that the URD requirements I just posted are aparently not requirements at all, so we'll likely be (re)moving them in the URD update. For this PR however that means I'll be reverting the look of the measure tool entirely to how it used to look. Most of the other things we discussed are inline with their reply.

RoanH:
Short update on the rendering. While it is possible to use a cairo transform to make it so post renderers can use presentation space coordinates to draw. It is not sufficient to just do this using a translation (unless the zoom stays at 1). To properly handle zooming it would be required to also scale everything, however, while this would make the coordinates convert properly, this also causes all drawing done by post renderers to get scaled which is not exactly desired. So for now I'll just go with the other option of passing all the relevant information to the post renderer.

gui/src/view.cc
```diff
@@ -185,23 +191,18 @@ void View::redraw(cairo_t* cr)
...
-      cairo_set_line_width(cr, 1);
-      cairo_set_source_rgb(cr, 0.75, 0, 0); // Dark Red
-      drawCross(cr, start);
-      drawCross(cr, end);
-      cairo_stroke(cr);
-      cairo_set_source_rgb(cr, 1, 0, 0); // Red
-      cairo_move_to(cr, start.x, start.y);
-      cairo_line_to(cr, end.x, end.y);
-      cairo_stroke(cr);
+      renderer->render(shared_from_this<View>(), cr, rect, zoom);
```
kees-jan, line 205: This feels inconsistent.

`PresentationInterface::redraw()` receives a `rect` argument, the idea being that it should draw whatever that rectangle contains at location `0,0` of the surface.

`PostRenderer::redraw()` receives the same `rect` argument, but this time the idea is that whatever that rectangle contains should be drawn at its "proper" location.

gui/src/view.cc
```diff
@@ -785,6 +774,78 @@ void View::removeFromToolbar(GtkToolItem* ti)
...
+  {
+    listener->onEnable();
+  }
+  else
+  {
+    listener->onDisable();
+  }
+}
+
+void View::addToolButton(const std::string& name, ToolStateListener::Ptr callback)
```
kees-jan, line 820: You really want to add a GtkWidget or button or whatever here. Names are ugly and should (in time) be replaced by icons, which should be under control of the plugin

gui/src/view.hh
```diff
@@ -79,10 +64,14 @@ private:
   int zoom;
   int x;
   int y;
-  Measurement* measurement;
+  Selection::Ptr selection;
+  std::vector<SelectionListener::Ptr> selectionListeners;
+  std::vector<PostRenderer::Ptr> postRenderers;
+  GtkWidget* pluginButtonGroup;
```
kees-jan, line 70: `toolButtonGroup` ?

inc/scroom/viewinterface.hh
```diff
@@ -7,14 +7,156 @@
...
 #include <boost/weak_ptr.hpp>
 
 #include <scroom/progressinterface.hh>
+#include <scroom/stuff.hh>
+#include <scroom/rectangle.hh>
+
+/**
+ * Forward declaration for ViewInterface::getCurrentPresentation()
+ */
+class PresentationInterface;
```
kees-jan, line 27: I propose to get rid of comments on forward declarations.
- The declaration itself is clear enough
- Because of the `/**`, It'll end up n the doxygen api documentation, which is definitely not desirable

inc/scroom/viewinterface.hh
```diff
@@ -7,14 +7,156 @@
...
+{
+public:
+  typedef boost::shared_ptr<ToolStateListener> Ptr;
+
+public:
+  virtual ~ToolStateListener() {}
+
+  /**
+   * This function is called whenever the tool
+   * button is select.
```
kees-jan, line 150: select_ed_?

gui/src/view.cc
RoanH: Fair enough, mostly did this due to the earlier comments, just reverted it. They both now just get the `rect` the translation for post renderers is gone.

gui/src/view.cc
RoanH: Fair enough, this does pose a bit of a problem though, but perhaps this is only because of my limited experience with GTK+2.

- As far as I could tell the only button types available in GTK+2 that support button groups are the `GtkRadioButton` and `GtkRadioToolButton`. 
- As far as I can tell from the documentation there is no way to create a new `GtkRadioButton` with some other `GtkWidget` in/next to it.
- I did not manage to get any custom icons to show the normal way in a `GtkRadioButton`. 
- I did manage to get an icon/label to show in a `GtkRadioToolButton`, however this functionality seems to be limited to stock items in GTK+2 and they are size wise rather large (see image) in addition it's not really that clear when it is toggled, though that might have been because I did not hook it up to anything.
afbeelding

Assuming that I am not missing something here then the only alternative implementations I can think of are:
- Adding both a `GtkRadioButton` and some `GtkWidget` supplied by the plugin on each tool button add which feels a bit hacky. 
- The `GtkRadioToolButton` seems promising, but also really limited (on first sight anyway). In addition stock items were deprecated in GTK3 with seemingly no replacement. 
- It would also technically be possible to implement our own button group logic if some other button type is more promising.

If you have anything in mind that I missed that allows both plugins to control what is shown and the view to managed the mutually exclusive aspect then please let me know @kees-jan. As it stands, I'm not entirely sure what the best course of action would be.

RoanH: I should also note that radio buttons with text don't really look all that great either due to it being hard to tell which radio button belongs to which label. Though I wasn't really able to find any better alternatives (so far).

inc/scroom/viewinterface.hh
```diff
@@ -234,16 +228,12 @@ public:
...
-  }
+  virtual void setPanning()=0;
 
   /**
    * Disable panning the view.
    */
-  virtual void unsetPanning()
-  {
-  }
+  virtual void unsetPanning()=0;
```
kees-jan, line 236: Now that selections are done with the right mouse button, are these still necessary?

plugins/measure/src/measure.cc
```diff
@@ -130,8 +130,8 @@ void MeasureHandler::render(ViewInterface::Ptr const& vi, cairo_t* cr, Scroom::U
 
   if(selection)
   {
-    Scroom::Utils::Point<double> start(selection->start.x, selection->start.y);
-    Scroom::Utils::Point<double> end(selection->end.x, selection->end.y);
+    Scroom::Utils::Point<double> start(selection->start.x - presentationArea.getLeft(), selection->start.y - presentationArea.getTop());
+    Scroom::Utils::Point<double> end(selection->end.x - presentationArea.getLeft(), selection->end.y - presentationArea.getTop());
```
kees-jan, line 134: the `Point` class probably needs a constructor that takes a `GdkPoint`.

Prefer doing arithmetic with entire point, instead of individual coordinates. So, eh...
```
auto start = Point(selection->start) - presentationArea.getTopLeft();
```

plugins/transparent-overlay/src/transparentoverlayviewinfo.cc
```diff
@@ -73,6 +73,29 @@ void ChildView::addToToolbar(GtkToolItem*)
...
+
+void ChildView::registerPostRenderer(PostRenderer::Ptr)
+{}
+
+void ChildView::setStatusMessage(const std::string&)
+{}
+
+boost::shared_ptr<PresentationInterface> ChildView::getCurrentPresentation()
+{
+  return nullptr;
```
kees-jan, line 93: This could now return the correct child presentation, right?

gui/src/view.cc
kees-jan: Didn't really dive into the technical details, but I'd say we're aiming for a look very similar to gimp, since that's what users will be expecting
image

They manage to make it clear which tool is selected, and use non-stock icons. Fortunately, the code is available :wink:

RoanH: That's a good idea, I'll look into the GIMP source, hopefully they did not implement something custom for this.

inc/scroom/viewinterface.hh
RoanH: I don't think there's a use case left for them and judging by your comment I guess you'd prefer them gone so I'll remove them later. If for some unlikely reason we end up needing them again then adding them back isn't a lot of effort either.

plugins/measure/src/measure.cc
RoanH: As someone who usually doesn't write c++ I always forget that operator overrides are a thing... Will change this in a bit.

kees-jan:
You did notice that you have a [cppcheck violation](https://app.codacy.com/manual/kees-jan/scroom/pullRequest?bid=18389439&prid=5613853&utm_campaign=new_commit&utm_medium=Email&utm_source=Internal&), right?

Also, [your builds](https://travis-ci.org/github/kees-jan/scroom/pull_requests) have a habit of failing

plugins/measure/src/measure.cc
kees-jan: Even if operator overrides weren't a thing, I'd probably still recommend you write a function that adds two points, instead of doing it manually. The rest is (unimportant) syntactic sugar

RoanH:
> 
> 
> You did notice that you have a [cppcheck violation](https://app.codacy.com/manual/kees-jan/scroom/pullRequest?bid=18389439&prid=5613853&utm_campaign=new_commit&utm_medium=Email&utm_source=Internal&), right?

I did not notice the cppcheck violation and wasn't aware of the automatic checks for that either, though I'll keep it in mind and fix the issue that's currently there.

> Also, [your builds](https://travis-ci.org/github/kees-jan/scroom/pull_requests) have a habit of failing

I actually have no idea why the builds started failing starting from your commit that changes measure to work in development mode, which really shouldn't have broken anything as far as I can tell. 

Now I just noticed that the reason CI failed has changed, currently it seems that I made some mistake with the `addToolButton` change that also causes the Linux CI to fail, though since I'll be changing that part soon anyway it'll either resolve itself or I'll look into it.

Before only the Windows build was failing with the only error I could find being `ERROR: The process "gpg-agent.exe" not found.`. Which I was expecting a merge of `next` into this branch to maybe fix, but I was putting that off until most of the other changes were in incase some other PR got merged in the mean time.

RoanH:
I should've addressed everything except for the tool buttons which I'll look at again later. Also merged next to see if the changes made to the Travis config affect the Windows CI status.

gui/src/view.cc
RoanH: Seems like GIMP both has some nice wrappers for this and is on GTK3 where there's a constructor for buttons with an image.

I did however finally get button images to work (turns out labels override images even if the image is set after the label).
afbeelding
So my plan for now is allow plugins to pass a `GtkToggleButton` or any of its sub classes and then implement button grouping manually. A plugin could then either return a toggle button with a text label or an image. Let me know if you think this is an acceptable solution.

RoanH: Pushed a commit for this. I'm currently not using an image for the measure tool because I don't really have a good image, but assuming it's not difficult to load/embed a png file (or raw image data I guess) in the plugin binary this wouldn't be too hard to add. I found that 16x16 pixels works best display wise.

There's also one last small 'issue' left. Currently disabling a plugin doesn't clear the status bar. I'm not sure whether to consider this a feature, responsibility of the plugin to clean up or responsibility of the view to clean up.

RoanH:
I think I addessed everything now, currently trying to figure out why Travis fails.

RoanH:
I verified that a `make distcheck` indeed [fails](https://gitlab.com/sep-group-2/scroom/-/jobs/586146823) to run on this branch. So I tried reverting the first commit that made Travis fail on this branch 12b143aa94619bdaba3adf6f04b5fd506bbcdf11. This does in fact make it so that the `make distcheck` [succeeds](https://gitlab.com/sep-group-2/scroom/-/jobs/586231766). I don't understand though why adding the measure plugin to `ENVIRONMENT.in` would cause this. Is there perhaps something else to take into account when it comes to the `Makefile.am` or `configure.ac` that I forgot to add?

gui/src/view.cc
kees-jan: Excellent. Wouldn't worry (at least for now) about text remaining in the statusbar.

kees-jan:
Did some testing, it looks good. A few minor things:
- I would expect the first toolbutton to register to be automatically enabled
- I did not expect being able to disable a tool (except by clicking another)

Currently experimenting with travis builds. I'm currently unable to reproduce the failure locally.

RoanH:
> Did some testing, it looks good. A few minor things:
> 
>     * I would expect the first toolbutton to register to be automatically enabled

Sure, I wasn't sure whether essentially enabling a random tool would be a good idea, but I'll change this.

>     * I did not expect being able to disable a tool (except by clicking another)

I could enable a tool button again immediately after it was clicked (and not forward the event to the plugins), I'm not sure if this poses any issues, so I'll try it out.

> Currently experimenting with travis builds. I'm currently unable to reproduce the failure locally.

Alright, I'll wait for your results then for now as I don't really want to spam Travis with 'fix' attempts when I have no idea about the underlying issue.

kees-jan:
> > Currently experimenting with travis builds. I'm currently unable to reproduce the failure locally.
> 
> Alright, I'll wait for your results then for now as I don't really want to spam Travis with 'fix' attempts when I have no idea about the underlying issue.

I just pushed a change to next that hopefully fixes the build problems. Still no clue what's causing it, though...

RoanH:
I changed it so the first tool registered has its button enabled. The position of the check for this is at the end of the function because a future plugin might want to add other tool bar items when it is enabled (and remove them when its disabled). This way the plugin tool button will always stay before any other tool bar items it might add. I'm not sure whether the `onEnable` call should be made from within or outside the gdk threads enter/leave block. From other UI frameworks I'm used to event handlers running on the UI thread, but I couldn't find a clear answer for GTK+2.

As for preventing a tool from getting disabled. The 'toggled' event is fired as well when the button state is programmatically updated, this causes an annoying recursive cycle that ends with a segfault when the stack overflows... I currently don't have any ideas for a simple solution to this problem.

> I just pushed a change to next that hopefully fixes the build problems. Still no clue what's causing it, though...

Weird... I'll merge next into this branch again then.

kees-jan:
>  I'm not sure whether the `onEnable` call should be made from within or outside the gdk threads enter/leave block. From other UI frameworks I'm used to event handlers running on the UI thread, but I couldn't find a clear answer for GTK+2.

On GTK2, event handlers are running on the UI thread too. However, changes to the UI must also be done from the UI thread (or at least holding the lock), so in this context, you can assume that the plugin takes care of this.

> As for preventing a tool from getting disabled. The 'toggled' event is fired as well when the button state is programmatically updated, this causes an annoying recursive cycle that ends with a segfault when the stack overflows... I currently don't have any ideas for a simple solution to this problem.

How about not listening to the 'toggled' event, but to the 'clicked' event instead?

RoanH:
> > I'm not sure whether the `onEnable` call should be made from within or outside the gdk threads enter/leave block. From other UI frameworks I'm used to event handlers running on the UI thread, but I couldn't find a clear answer for GTK+2.
> 
> On GTK2, event handlers are running on the UI thread too. However, changes to the UI must also be done from the UI thread (or at least holding the lock), so in this context, you can assume that the plugin takes care of this.

Alright, I'll leave this as is then.

> > As for preventing a tool from getting disabled. The 'toggled' event is fired as well when the button state is programmatically updated, this causes an annoying recursive cycle that ends with a segfault when the stack overflows... I currently don't have any ideas for a simple solution to this problem.
> 
> How about not listening to the 'toggled' event, but to the 'clicked' event instead?

Turns out that the `clicked` event also fires when programmatically updating the button. However the `pressed` and `released` events do not fire. They have their own issues though:
-  The `pressed` event fires too early making it impossible to cancel the press. In addition it seems to fire before the toggle is handled making it so updating the button state in the handler desyncs the internal state from what's actually shown on screen resulting in a visually unpressed toggle button with a pressed internal state.
- The `released` event could work, but it's possible to drag the mouse outside of the button bounds and then release it. That could be solved if it's easy to get the current mouse position and check if that's within the button bounds though. It also has a similar issue to the `pressed` event where the visual and internal state desync, but here it's only for a split second and not permanent. I guess it might be an option to listen for the enter/leave events for each button as an alternative to checking if the current mouse position is within bounds. Either way a solution using this signal would be quite hacky I feel.
- I also tried the `button-release-event` signal, but this one just freezes the UI and I feel like I'm way too far up the class hierarcy at this point.

For now I'm out of ideas again...

RoanH:
Found a decent alternative. We can just disable the button of the active tool, this both makes it visually really clear to the user that there's no point clicking on the button again and also stops any mouse events from registering on the button so no events get fired. Then when a different tool is enabled we just disable the button for that tool and re-enable the button for the previous tool. On Linux it would look like this:
afbeelding

PR #60, opened 2020-06-03
Pixel averages

ArminZahedani:
Hi @kees-jan, 
This is the second PR about the Pipette Tool, which is mostly about getting the average pixel values, given a selection (area) of a presentation. I will summarize the changes below:

### Changes:
1. We implemented a new interface `PipetteViewInterface` which declares a method `getAverages()`. This method will get implemented in `TiffPresentation`. The function takes an area as input, which is an area that is contained in the presentation. It will then return the averages (mean) of the pixel components in the area. The case for a selection outside the presentation is handled outside by the pipette tool right now.
2. We implemented a new interface `PipetteLayerOperations` which declares a method `sumPixelValues()`. This method will, given an area and a tile, where the area is completely contained in the tile, compute the sum of all pixel components. The class also defines `PipetteColor` which is a map of strings and size_t. 
3. All OperationsCMYK (4, 8, 16, 32) and (for now) `Operations24bpp` inherit from both `CommonOperations` and `PipetteLayerOperations`.
4. `TiffPresentation` has a method `getAverages()`. The `TiffPresentationWrapper` inherits from the `PipetteViewInterface` and delegates the call to the `getAverages()` method in `TiffPresentation`. 
5. The implementation of `getAverages()` first dynamically casts to make sure that the `PipetteLayerOperations` are implemented. If not, an empty map is returned. If the `getAverages()` method is supported, the method computes which tiles intersect with the selected area, and then loops over them. It the grabs the tile and defines a sub rectangle that is completely defined in the tile, and its coordinates are adjusted to be relative to the tile.
6. The specific implementation for `sumPixelValues()` is then called with this sub rectangle and the corresponding tile. The map that gets returned is then added using the `Operator+=` method and at the end divided by the totalPixels with the `Operator/` method
7. The implementation and specific details for `sumPixelValues()` for each Operations loops over the rectangle and extracts the values for each channel, adds them and returns them in a map.

### Considerations:

1. As said in Point 3, the Operations inherit from `CommonOperations` and `PipetteLayerOperations`. The original design that we discussed combined the two interfaces into a class `PipetteCommonOperations`. We moved away from this idea for now, given that we implemented operations for specific cases e.g. CMYK 8bps, 4bps etc.  what do you think about the current implementation and do you think there are any reasons to move back to the original idea?
2. We have defined two custom operator methods. `Operator+=` and `Operator/` for `PipetteColor`. We are not sure where to place them. We could also include the methods in `PippeteLayerOperations.hh` using `friend`. Any suggestions there?
3. To access the `LayerOperations` from the presentation, we made `Layerspec ls` a member variable of `TiffPresentation`.
4. The operations for `sumPixelValues()` have not been implemented for `Operations1bpp`, `Operations1bppClipped` and `Operations8bpp` right now. At this moment we are not completely sure how to implement using the colormaps.

We tried our best to keep as close as possible to the coding style. If there are any other comments, we are happy to hear. 
Regards,
Armin


ArminZahedani:
> * `SumPixelValues()` should probably return a `vector<size_t>` as a result, because you'll want to change `PipetteColor` to map to doubles, to support returning fractional samples. Also, it'll be faster to add two vectors than two maps 😉
> * I noticed that you still are not a big fan of `SampleIterator`s. However, you have now code that extracts samples from bitmap data in `cache()`, `reduce()` and `sumPixelValues()`, and that over 4 different classes. Is there anything we can do here to reduce duplication?
> * Have presentations return a property, to indicate whether they support the pippette tool. That way
>   
>   * No pippette tool button will be shown for presentations that don't support it
>   * You'll never have to return an empty `PipetteColor`
>   * All casts to Pipette related interfaces should succeed, so you can replace a bunch of `if`s by `assert`s.

The reason for the first one is the presentation would then need to check whether its CMYK or RGB (probably based on the length of the vector). This felt easier to do with the map and also allows us to just re-use the keys when displaying the values in the status bar. I feel like this way would be a bit safer but i don't have a preference for this.

The second point is a fair point. I think for `SampleIterator` there was some confusion going on with how to use it, and hence we decided not to use it. Given the example's you provided, we will have a look into the use of the `SampleIterator` in this use case. 

I very much like the third suggestion ^^. I guess this is similar to e.g. `COLORMAPPABLE_PROPERTY_NAME` in `tiffpresentation.cc` using `properties`?

plugins/tiff/src/tiffpresentation.cc
```diff
@@ -389,6 +390,85 @@ void TiffPresentation::done()
...
+  }
+  return x;
+}
+
+////////////////////////////////////////////////////////////////////////
+// PipetteViewInterface
+////////////////////////////////////////////////////////////////////////
+PipetteLayerOperations::PipetteColor TiffPresentation::getAverages(Scroom::Utils::Rectangle<int> area)
+{
+  PipetteLayerOperations::Ptr pipetteLayerOperation = boost::dynamic_pointer_cast<PipetteLayerOperations>(this->ls[0]);
```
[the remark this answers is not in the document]
ArminZahedani: `ls` was made a member as we thought we may be able to make use of the reduced layer on top of the most bottom layer. However, this idea was not implemented for now. Indeed, we can make `PipetteLayerOperation` a member instead, which is probably better ^^.

inc/scroom/pipettelayeroperations.hh
```diff
@@ -0,0 +1,30 @@
...
+public:
+  typedef boost::shared_ptr<PipetteLayerOperations> Ptr;
+  typedef std::map<std::string, size_t> PipetteColor;
+
+public:
+  virtual ~PipetteLayerOperations()
+  {}
+
+  /**
+   * Method that each layer Operation that supports the pipette tool will implement.
```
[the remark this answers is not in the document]
ArminZahedani: Yes, the comments will be improved ^^

inc/scroom/pipetteviewinterface.hh
```diff
@@ -0,0 +1,28 @@
...
+public:
+  virtual ~PipetteViewInterface()
+  {}
+
+  /** 
+   * Method to get the average pixel values in the area.
+   * The method assumes that the selected area is completely contained in the presentation.
+   * Returns a map of strings and size_t, representing averages for each component.
+   */
+  virtual PipetteLayerOperations::PipetteColor getAverages(Scroom::Utils::Rectangle<int> area)=0;
```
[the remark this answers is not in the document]
ArminZahedani: what about `getPixelAverages()` ?

plugins/tiff/src/tiffpresentation.cc
```diff
@@ -389,6 +390,85 @@ void TiffPresentation::done()
...
+ * Returns y if x is empty.
+ * Otherwise adds the values of the same key.
+ */
+PipetteLayerOperations::PipetteColor operator+=(PipetteLayerOperations::PipetteColor& x, const PipetteLayerOperations::PipetteColor& y)
+{
+  if(x.empty())
+  {
+    x = y;
+    return x;
+  }
```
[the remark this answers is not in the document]
ArminZahedani: This will be (was already) reverted ^^

inc/scroom/pipetteviewinterface.hh
```diff
@@ -0,0 +1,28 @@
...
+public:
+  typedef boost::shared_ptr<PipetteViewInterface> Ptr;
+
+public:
+  virtual ~PipetteViewInterface()
+  {}
+
+  /** 
+   * Method to get the average pixel values in the area.
+   * The method assumes that the selected area is completely contained in the presentation.
```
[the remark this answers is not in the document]
ArminZahedani: I guess this can be solved with using `intersection()`. Will be fixed ^^

plugins/tiff/src/tiffpresentation.cc
```diff
@@ -389,6 +390,85 @@ void TiffPresentation::done()
...
+
+  //Get end tile (tile_pos_x_end, tile_pos_y_end)
+  int tile_pos_x_end = (area.getRight() - 1) / TILESIZE;
+  int tile_pos_y_end = (area.getBottom() - 1) / TILESIZE;
+
+  for(int x = tile_pos_x_start; x <= tile_pos_x_end; x++)
+  {
+    for(int y = tile_pos_y_start; y <= tile_pos_y_end; y++)
+    {
+      ConstTile::Ptr tile = bottomLayer->getTile(x, y)->getConstTileSync();
```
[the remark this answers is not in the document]
ArminZahedani: I am not sure if this is a priority at this point for us (given our limited time etc.). For sure blocking the UI thread is not the best approach. We will see whether we will work on this in a future PR...

ArminZahedani:
> I noticed that you still are not a big fan of SampleIterators. However, you have now code that extracts samples from bitmap data in cache(), reduce() and sumPixelValues(), and that over 4 different classes. Is there anything we can do here to reduce duplication?

A small update on using Sample iterators:
Using `SampleIterator` indeed makes the code a lot nicer. However, each `OperationsCMYKx` needs a different `SampleIterator` which creates a slightly different initialization of the `SampleIterator`. We could introduce a method that initializes the sampleiterator and then call `sumPixelValues()` with the correct sample iterator settings. Do you have any preference for this?

Furthermore, I will revert the change of making `ls` a member of `TiffPresentation` and I am thinking about introducing `LayerOperations::Ptr pipetteLayer` as a member instead. This will then grab `ls[0]` whenever the operations has been added to `ls`. Is this fine?

inc/scroom/pipetteviewinterface.hh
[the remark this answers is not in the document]
kees-jan: Better :wink:

plugins/tiff/src/tiffpresentation.cc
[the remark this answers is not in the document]
kees-jan: I believe there are some performance requirements in the URD :wink:

inc/scroom/layeroperations.hh
```diff
@@ -35,7 +36,7 @@ private:
   ColormapProvider::Ptr colormapProvider;
 
 public:
-  static Ptr create(ColormapProvider::Ptr colormapProvider);
+  static LayerOperations::Ptr create(ColormapProvider::Ptr colormapProvider);
```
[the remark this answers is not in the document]
kees-jan: Ambiguity is probably with `PipetteLayerOperations::Ptr`, which is currently unused. So you might as well delete that `typedef`.

plugins/tiff/src/tiffpresentation.cc
```diff
@@ -389,6 +390,85 @@ void TiffPresentation::done()
   tif = NULL;
 }
 
+/**
+ * Method to override operator+= method
```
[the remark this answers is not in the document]
kees-jan: Thing is that we are operating on a `map<string,size_t>` or something, which is a fairly generic type. And then the `operator+=` is a fairly generic method. But the implementation does something extremely specific: Adding color sample values.

For an operation so specific, it is best to use either a specific type, or a specific method name. The current implementation leads to confusion too easily

kees-jan:
> A small update on using Sample iterators:
> Using `SampleIterator` indeed makes the code a lot nicer. However, each `OperationsCMYKx` needs a different `SampleIterator` which creates a slightly different initialization of the `SampleIterator`. We could introduce a method that initializes the sampleiterator and then call `sumPixelValues()` with the correct sample iterator settings. Do you have any preference for this?

Only difference in initialization should be the `bps` value. You could pass that in (assuming the `LayerOperations` implementation doesn't yet know that itself. Passing in a completely initialized `SampleIterator` seems to make less sense to me.

> Furthermore, I will revert the change of making `ls` a member of `TiffPresentation` and I am thinking about introducing `LayerOperations::Ptr pipetteLayer` as a member instead. This will then grab `ls[0]` whenever the operations has been added to `ls`. Is this fine?

To avoid casts, you'll want that to be a `PipetteLayerOperations::Ptr`. That means that grabbling `ls[0]` might not work any more (on presentations that don't support pippete), so I would probably initialize `ls[0]` and that new pointer at the exact same time, in code where you know the presentation will support pipette.

inc/scroom/layeroperations.hh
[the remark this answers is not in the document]
ArminZahedani: I am not sure whether this can be removed. Removing this would also change the way in which the `dynamic_pointer_cast` is done at the top of the implementation of `getPixelAverages()`. If we remove this cast (which i am not really sure if we can), we can probably remove that `typedef`

plugins/tiff/src/tiffpresentation.cc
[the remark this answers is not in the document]
ArminZahedani: Indeed, that is a fair point. I changed this now ^^

ArminZahedani:
> > A small update on using Sample iterators:
> > Using `SampleIterator` indeed makes the code a lot nicer. However, each `OperationsCMYKx` needs a different `SampleIterator` which creates a slightly different initialization of the `SampleIterator`. We could introduce a method that initializes the sampleiterator and then call `sumPixelValues()` with the correct sample iterator settings. Do you have any preference for this?
> 
> Only difference in initialization should be the `bps` value. You could pass that in (assuming the `LayerOperations` implementation doesn't yet know that itself. Passing in a completely initialized `SampleIterator` seems to make less sense to me.

In that case, we could re-introduce `PipetteCommonOperations` and add `bps` and `spp` to the `sumPixelValues()` method. We then implement the method on `PipetteCommonOperations` only, checking whether the `spp` is 3 or 4 (for RGB vs CMYK) and using the bps for the `SampleIterator` .

> > Furthermore, I will revert the change of making `ls` a member of `TiffPresentation` and I am thinking about introducing `LayerOperations::Ptr pipetteLayer` as a member instead. This will then grab `ls[0]` whenever the operations has been added to `ls`. Is this fine?
> 
> To avoid casts, you'll want that to be a `PipetteLayerOperations::Ptr`. That means that grabbling `ls[0]` might not work any more (on presentations that don't support pippete), so I would probably initialize `ls[0]` and that new pointer at the exact same time, in code where you know the presentation will support pipette.

I cannot really follow here. In combination with my comment from above about the `dynamic_pointer_cast` i am not sure whether a cast can be avoided here, but i will look into this more

plugins/tiff/src/tiffpresentation.cc
```diff
@@ -389,6 +390,85 @@ void TiffPresentation::done()
...
+
+  int totalPixels = area.getWidth() * area.getHeight();
+  
+  if(totalPixels == 0){
+    return {};
+  }
+
+  //Get start tile (tile_pos_x_start, tile_pos_y_start)
+  int tile_pos_x_start = (area.getLeft() - 1) / TILESIZE;
+  int tile_pos_y_start = (area.getTop() - 1) / TILESIZE;
```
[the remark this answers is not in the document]
ArminZahedani: The `-1` for the `area.getLeft()` and `area.getTop()` was a mistake from our side. It is needed for `area.getBottom()` and `area.getRight()` though, as they appear to return the pixel one to the bottom/right of the selected area.

inc/scroom/transformpresentation.hh
```diff
@@ -60,7 +60,12 @@ public:
   virtual std::string getTitle();
 
   // PipetteViewInterface
-  virtual PipetteLayerOperations::PipetteColor getAverages(Scroom::Utils::Rectangle<int> area);
+  /** 
+   * Returns the average pixel values for each component, contained in the area
+   * 
+   * @param area selected area to get the pixel values from.
+   */
+  virtual PipetteLayerOperations::PipetteColor getPixelAverages(Scroom::Utils::Rectangle<int> area);
```
kees-jan, line 68: When you implement an interface, do not repeat the documentation.

I think, doxygen is smart enough to look up the documentation of the base class. And even if not, repeating the same comment over and over in the code is begging for trouble if they ever need to be updated.

libs/scroom/src/transformpresentation.cc
```diff
@@ -114,12 +114,12 @@ bool TransformPresentation::getTransparentBackground()
   return colormappable->getTransparentBackground();
 }
 
-PipetteLayerOperations::PipetteColor TransformPresentation::getAverages(Scroom::Utils::Rectangle<int> area)
+PipetteLayerOperations::PipetteColor TransformPresentation::getPixelAverages(Scroom::Utils::Rectangle<int> area)
 {
-  PipetteViewInterface::Ptr pipetteLayer = boost::dynamic_pointer_cast<PipetteViewInterface>(presentation);
-  if(pipetteLayer)
+  PipetteViewInterface::Ptr pipettePresentation = boost::dynamic_pointer_cast<PipetteViewInterface>(presentation);
+  if(pipettePresentation)
```
kees-jan, line 120: That should be `require(pipettePresentation)`: I.e. if someone calls this, then the cast must succeed (caller should inspect the correct property on the presentation to guarantee this)

There's a few other places where this can be improved.

plugins/tiff/src/tiffpresentation.cc
```diff
@@ -195,28 +195,39 @@ bool TiffPresentation::load(const std::string& fileName_)
     printf("This bitmap has size %d*%d, aspect ratio %.1f*%.1f\n",
            width, height, 1/resolutionX, 1/resolutionY);
 
+    LayerSpec ls;
     if (spp == 4 && bps == 8)
     {
       ls.push_back(OperationsCMYK32::create());
+      properties[PIPETTE_PROPERTY_NAME] = "";
+      this->pipetteLayer = boost::dynamic_pointer_cast<PipetteLayerOperations>(ls[0]);
```
kees-jan, line 203: No need for the `this->` (this isn't java :wink: )

By changing the order around, you can avoid the dynamic cast:
```
pipetteLayer = OperationsCMYK32::create();
ls.push_back(pipetteLayer);
properties[PIPETTE_PROPERTY_NAME] = "";
```

... well... That would have worked if `PipetteLayerOperations` actually inherited from `LayerOperations`. Perhaps you should reconsider :wink:

plugins/tiff/src/tiffpresentation.cc
```diff
@@ -391,12 +402,12 @@ void TiffPresentation::done()
...
 
 /**
- * Overriden operator+= to add two values of the same key.
+ * Add two pipette color map values of the same key.
  * 
  * Loops through rhs, adds values of the same key to each other and assigns it back to lhs.
  * Return lhs at the end.
  */
-PipetteLayerOperations::PipetteColor operator+=(PipetteLayerOperations::PipetteColor& lhs, const PipetteLayerOperations::PipetteColor& rhs)
+PipetteLayerOperations::PipetteColor sumPipetteColors(PipetteLayerOperations::PipetteColor& lhs, const PipetteLayerOperations::PipetteColor& rhs)
```
kees-jan, line 410: You are taking `lhs` by reference, and then returning a non-reference.

I recommend making `lhs` a const reference. Callers may not expect you to modify it.

plugins/tiff/src/tiffpresentation.cc
```diff
@@ -406,11 +417,9 @@ PipetteLayerOperations::PipetteColor operator+=(PipetteLayerOperations::PipetteC
...
 /**
- * Overriden operator/ method to divide each element in the map by a constant.
- * 
  * Divides each element inside elements by by a constant divisor.
  */
-PipetteLayerOperations::PipetteColor operator/(PipetteLayerOperations::PipetteColor elements, const int divisor )
+PipetteLayerOperations::PipetteColor dividePipetteColors(PipetteLayerOperations::PipetteColor elements, const int divisor )
 {
   for(auto elem : elements)
   {
```
kees-jan, line 425: There's an unnecessary lookup in the map here. How about
```
elem.second /= divisor;
```

plugins/tiff/src/tiffpresentation.cc
```diff
@@ -422,14 +431,18 @@ PipetteLayerOperations::PipetteColor operator/(PipetteLayerOperations::PipetteCo
 ////////////////////////////////////////////////////////////////////////
 // PipetteViewInterface
 ////////////////////////////////////////////////////////////////////////
-PipetteLayerOperations::PipetteColor TiffPresentation::getAverages(Scroom::Utils::Rectangle<int> area)
+PipetteLayerOperations::PipetteColor TiffPresentation::getPixelAverages(Scroom::Utils::Rectangle<int> area)
 {
-  PipetteLayerOperations::Ptr pipetteLayerOperation = boost::dynamic_pointer_cast<PipetteLayerOperations>(this->ls[0]);
+  PipetteLayerOperations::Ptr pipetteLayerOperation = this->pipetteLayer;
```
kees-jan, line 436: Now replace `pipetteLayerOperation` with `pipetteLayer` (no `this->`) everywhere :wink:

plugins/tiff/src/tiffpresentation.cc
```diff
@@ -422,14 +431,18 @@ PipetteLayerOperations::PipetteColor operator/(PipetteLayerOperations::PipetteCo
...
+
   if(!pipetteLayerOperation)
```
kees-jan, line 438: This, too, should be a `require()`

plugins/tiff/src/tiffpresentation.hh
```diff
@@ -36,14 +36,15 @@ private:
...
   TiledBitmapInterface::Ptr tbi;
-  LayerSpec ls;
   int bps;
   int spp;
   std::map<std::string, std::string> properties;
   Views views;
   ColormapHelper::Ptr colormapHelper;
   TransformationData::Ptr transformationData;
 
+  PipetteLayerOperations::Ptr pipetteLayer;  //holds the most bottom layer for pipette operations
```
kees-jan, line 46: Actually, it doesn't hold the most bottom layer, but the most bottom `LayerOperations`.

So perhaps you should remove the comment and rename the member variable?

plugins/tiff/src/tiffpresentation.cc
```diff
@@ -389,6 +390,85 @@ void TiffPresentation::done()
...
+  {
+    return {};
+  }
+
+  Layer::Ptr bottomLayer = this->tbi->getBottomLayer();
+  PipetteLayerOperations::PipetteColor pipetteColors;
+
+  int totalPixels = area.getWidth() * area.getHeight();
+  
+  if(totalPixels == 0){
```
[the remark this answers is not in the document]
kees-jan: In case of an empty selection, `PixelColors` will be an empty map, and hence you will not enter the body of the for-loop that does the division.

plugins/tiff/src/tiffpresentation.cc
ArminZahedani: This is an interesting suggestion that I just tried out. Essentially this would at the current stage introduce a diamond hierarchy (`PipetteCommonOperations` would inherit from both `PipetteLayerOperations` and `CommonOperations`; Both `PipetteLayerOperations` and `CommonOperations` would inherit from `LayerOperations`). While this can work, i believe there are some problems which need to be fixed and i could not really identify the best approach (the Compiler is mostly complaining about some methods not being implemented in `OperationsCMYKx`/ `Operations24bpp`, probably because it is confused as to where it is implemented; The methods are `initializeCairo(), draw()` & `drawState()`):

1. Cut the link between `PipetteCommonOperations` and `CommonOperations`. This would mean that all methods defined in `CommonOperations` would need to be re-implemented in `PipetteCommonOperations`. This creates quite a lot of code duplication...
2. Re-declare the methods from `CommonOperations` in `PipetteCommonOperations` and delegate the call to `CommonOperations`. This would be my own preference, however does not feel that clean...
3. Restructure the whole hierarchy. Essentially, `CommonOperations` needs to inherit from `LayerOperations`, and we also want `PipetteLayerOperations`to inherit from `LayerOperations`. `PipetteCommonOperations` also needs to inherit from `PipetteLayerOperations`. The only link which I think could be cut would be from `PipetteCommonOperations` to `CommonOperations`. However, that would introduce code duplication...
4. Continue using the current solution with dynamic casts. They are not a great solution but perhaps we can reconsider these given the above problem.

These are the ideas that i currently have. Maybe you have a better solution for the above problem

kees-jan: Oops... That sounds more involved than I anticipated. I think of all of the above, I favor alternative 4 for simplicity.

There's one more variation I'd like you to consider
```
auto cmykOperations = OperationsCMYK32::create();
pipetteLayer = cmykOperations;
ls.push_back(cmykOperations);
properties[PIPETTE_PROPERTY_NAME] = "";
```
This is slightly longer than we had before, but has two upcasts, which are done implicitly, and can be compile-time-checked. As opposed to one cross-cast, which theoretically could fail at runtime.

inc/scroom/layeroperations.hh
```diff
@@ -29,6 +30,16 @@ public:
                     Scroom::Utils::Stuff cache);
 };
 
+class PipetteCommonOperations : public CommonOperations, public PipetteLayerOperations
+{
+public:
+  virtual ~PipetteCommonOperations()
+  {}
+
+  PipetteLayerOperations::PipetteColor sumPixelValues(Scroom::Utils::Rectangle<int> area, const ConstTile::Ptr tile, int bps, int spp);
```
kees-jan, line 39: But... but...

It is the job of the `LayerOperations` object to know the `spp` and `bps` values. Otherwise, it would never be able to do such things as `redraw()` and `reduce()`.

Hence, it should not be necessary to pass them in here.

libs/tiled-bitmap/src/layeroperations.cc
```diff
@@ -209,6 +209,52 @@ void CommonOperations::draw(cairo_t* cr, const ConstTile::Ptr tile,
...
+        sum_r += *si++;
+        sum_g += *si++;
+        sum_b += *si++;
+      }
+      si += spp * stride;
+    }
+  
+  return { {"R", sum_r}, {"G", sum_g}, {"B", sum_b} };
+  }
+  return {};
```
kees-jan, line 255: You guys really do insist on returning empty stuff, right :sob:

libs/tiled-bitmap/src/layeroperations.cc
```diff
@@ -209,6 +209,52 @@ void CommonOperations::draw(cairo_t* cr, const ConstTile::Ptr tile,
   }
 }
 
+PipetteLayerOperations::PipetteColor PipetteCommonOperations::sumPixelValues(Scroom::Utils::Rectangle<int> area, const ConstTile::Ptr tile, int bps, int spp)
```
kees-jan, line 212: If you are going to provide different implementation for `spp==4` and `spp==3`, then perhaps they should be in different classes. After all, they share virtually no code.

On the other hand, the only real difference between the two cases is the names of the samples. If `sumPixelValues()` were to return a vector of samples, instead of a map, then they would be identical.

Only problem left is to find a suitable location to convert the vector into a map :wink:

libs/tiled-bitmap/src/layeroperations.cc
```diff
@@ -209,6 +209,52 @@ void CommonOperations::draw(cairo_t* cr, const ConstTile::Ptr tile,
...
+{ 
+  Scroom::Bitmap::SampleIterator<const uint8_t> si(tile->data.get() + area.getTop(), 0, bps);
```
kees-jan, line 214: I doubt the computation of the starting point is correct. How about adding some tests?

plugins/tiff/src/tiffpresentation.cc
```diff
@@ -389,6 +401,85 @@ void TiffPresentation::done()
...
+ */
+PipetteLayerOperations::PipetteColor sumPipetteColors(const PipetteLayerOperations::PipetteColor& lhs, const PipetteLayerOperations::PipetteColor& rhs)
+{
+  PipetteLayerOperations::PipetteColor result;
+  for(auto const& elem : rhs)
+  {
+    result[elem.first] = elem.second;
+    if(lhs.find(elem.first) != lhs.end())
+    {
+      result[elem.first] += lhs.at(elem.first);
```
kees-jan, line 418: ... so first you copy `rhs` into result (in a roundabout way), and then, if `lhs` shares any key with `rhs`, you sum those values, but any key in `lhs` that is not shared by `rhs` is ignored.

That's quite different from what it did before. Why the change of heart?

plugins/tiff/src/tiffpresentation.cc
ArminZahedani: `OperationsCMYK32::create()` does return a `LayerOperations::Ptr` right now. How about changing this to `PipetteCommonOperations::Ptr` (and of course adding a typedef for the `PipetteCommonOperations`)?

kees-jan: Oeps... Yes, that would make sense, then.

libs/tiled-bitmap/src/layeroperations.cc
ArminZahedani: My bad. Indeed this is an error. I seem to have forgotten about scaling the vertical offset by `tile->width` and adding the horizontal offset. I think it should be `tile->data.get() + area.getTop() * tile->width + area.getLeft()`. Sorry about that...

plugins/tiff/src/tiffpresentation.cc
ArminZahedani: The problem for me was that if `lhs` is a `const`, we can not access elements using `[]` i.e. `lhs[element]` but have to use `.at()`. The issue here is then that this would throw an error when the key is not in the map, which is different to before where it would default to `0` if you used `[]` and the key was not in the map.

inc/scroom/layeroperations.hh
ArminZahedani: I guess we can add both `spp` and `bps` to `PipetteCommonOperations`  and define getters `getSpp` and `getBps` which each operation which supports pipette will implement?

Alternative would be to make them public in `PipetteCommonOperations` . If we split up the method `sumPixelValues` for CMYK vs RGB, we could get away with a hardcoded `spp` value and compute the `bps` from the `bpp`.

plugins/tiff/src/tiffpresentation.cc
kees-jan: But still. Why not
```
result  = lhs;
for(auto& [k,v] : rhs)
  result[k] = v;
```
or something similar?

inc/scroom/layeroperations.hh
kees-jan: I'm not a big fan of getters. I guess I'd make them protected. For now, we're only talking about PipetteCommonOperations, and since that's the class that implements `sumPixelValues()`, you might even make them private and not have any getters.

libs/tiled-bitmap/src/layeroperations.cc
ArminZahedani: I think for myself, i would prefer to split them into 2 classes. `PipetteCommonOperationsCMYK` and `PipetteCommonOperationsRGB` would be options imo. The specific operations such as `OperationsCMYK24` can then inherit from one of the two above (in this case being `PipetteCommonOperationsCMYK`). The spp would then be implicit, the bps *could* be calculated from the `getBpp()` function. (I think you noticed that im not a huge fan of vectors for this :smile: ). My only worry for this approach would be the class hierarchy and large inheritance chain that we would end up with for `OperationsCMYKx` and `Operationsxbpp`

plugins/tiff/src/tiffpresentation.cc
ArminZahedani: I have not thought about this 😅 . Note that this does throw a warning:
`warning: decomposition declaration only available with -std=c++1z or -std=gnu++1z`. I guess the `[k,v]` can be replaced by `elem`, and we can access the key and value using `elem.first` and `elem.second` respectively, like we did before.

plugins/tiff/src/tiffpresentation.cc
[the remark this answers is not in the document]
ArminZahedani: The performance requirements with not blocking the UI thread will be resolved in the next PR. The performance for the test images seems to meet the URD requirements so we are happy with that for now

libs/tiled-bitmap/src/layeroperations.cc
kees-jan: After all edits, the plan seems sound :wink: 

For future reference: You were asking me to review concrete `OperationCMYKx` classes, where some of the methods new the bpp and spp values they were supposed to be operating with, and some other methods were not. For me, that's a sign of bad design, and that was the trigger to start this discussion.

Now, a long inheritance hierarchy is "merely" a code smell. So although the concern is valid, introducing bad design to avoid a smell is, eh, misguided.

Having said all that, the inheritance tree we're creating here is still pretty much ok (imho, and stuff). Yes, there are a fair number of classes, but they have clear responsibilities, and all method implementations match the responsibilities of their respective classes. Furthermore, you'll notice that each method on the interface is implemented in exactly one location in the inheritance tree. There are no overrides, there are no calls to parent methods in base classes. We could do much worse. In fact, this tree is fairly easy to clean up, if the need arises. Maybe by moving from inheritance to composition. Maybe by splitting the `LayerOperations` interface further. There are several possibilities to choose from.

libs/tiled-bitmap/src/layeroperations.cc
ArminZahedani: Actually, there were some problems here with 1bps that i did not realize. These should be fixed now 😃

plugins/tiff/src/tiffpresentation.cc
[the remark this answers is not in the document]
ArminZahedani: I tested this again. If `totalPixels == 0` `pipetteColors` will not be an empty map, but a map of 0's. Then a division by 0 occurs which returns `-nans`.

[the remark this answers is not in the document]
kees-jan: You are saying that if totalPixels is 0, we're still going to decompress a tile, and call `sumPixelValues()` on it. Why? If we don't need any pixels of that particular tile?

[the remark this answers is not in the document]
ArminZahedani: Perhaps there is some miscommunication going on from my side. I meant that without the check and with `totalPixels == 0` the tile will get loaded. The check will make sure that if `totalPixels == 0` no tile will get loaded (as it will not reach the `for` loops).
 I am not sure if you are suggesting to take out the `if` check or if it is fine using this check. The check is needed because the `tile_pos_x_start` and `tile_pos_x_end` (and similarly for `y`)  cannot really distinguish whether a rectangle actually has an `area > 0` or not. They are just there to check to make sure to loop over the correct tiles.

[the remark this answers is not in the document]
kees-jan: Hmm... Perhaps it is easiest for now to leave the check in.

Yes, I was indeed suggesting that if the code that computes which tiles are needed, returns more tiles than are actually needed, then it contains a bug.

That's obviously the case when we have zero pixels, and yet the code returns the one tile :wink:

ArminZahedani:
I believe that all current concerns are handled now ^^

inc/scroom/layeroperations.hh
```diff
@@ -29,6 +30,42 @@ public:
                     Scroom::Utils::Stuff cache);
 };
 
+class PipetteCommonOperations : public CommonOperations, public PipetteLayerOperations
```
kees-jan, line 33: Not sure what the purpose of this empty class is?

inc/scroom/layeroperations.hh
```diff
@@ -69,11 +106,11 @@ public:
                     Scroom::Utils::Stuff cache);
 };
 
-class Operations24bpp : public CommonOperations
+class Operations24bpp : public PipetteCommonOperationsRGB
 {
 public:
-  static Ptr create();
-  Operations24bpp();
+  static PipetteCommonOperations::Ptr create(int bps);
```
kees-jan, line 112: For RGB 24bpp, bps is always 8, right? No need to pass it in

kees-jan: ... I see that something similar holds for the other concrete classes below.

inc/scroom/pipettelayeroperations.hh
```diff
@@ -0,0 +1,47 @@
...
+/**
+ * This string defines the order in which color channels are stored in the map
+ * and eventually in what order the channels will be displayed to the user.
+ */
+const std::string COLOR_CHANNELS = "CMYK" "RGB" "ADEFHIJLNOPQSTUVWXZ";
+
+/**
+ * This struct is just here to compare two keys according to the order defined above
+ */
+struct ColorComparator : public std::binary_function<std::string, std::string, bool> {
```
kees-jan, line 23: This breaks horribly when colors become more than one letter, as they will be for the "light" variants of CMYK

inc/scroom/layeroperations.hh
ArminZahedani: The reason for it was to 'group' the two `PipetteCommonOperationsRGB` and `PipetteCommonOperationsCMYK` when creating the `Ptr`'s of specific `LayerOperations` . If we don't want this class anymore, we would need to introduce  `typedef boost::shared_ptr<PipetteCommonOperationsRGB> Ptr;` (and similarly for CMYK). Then we can remove `PipetteCommonOperations`. Will be fixed

inc/scroom/layeroperations.hh
ArminZahedani: Do you mean something like:
```cpp
PipetteCommonOperationsCMYK::Ptr OperationsCMYK32::create()
{
  return PipetteCommonOperationsCMYK::Ptr(new OperationsCMYK32());
}

OperationsCMYK32::OperationsCMYK32() : PipetteCommonOperationsCMYK(8)
{
}
```
in e.g. `cmyklayeroperations.cc` and similarly for `layeroperations.cc`?

kees-jan: Yes, that would be it.

inc/scroom/pipettelayeroperations.hh
ArminZahedani: Light variants of CMYK channels would be sorted to the end right now. However, we can instead compare character by character in a lexicographic fashion, with our custom alphabet. CMYK does not support light variants right now.

kees-jan: Well, no, it is far worse than that. Assuming "LC", "LM" and "LY", your comparator will find none of them in the reference string. That causes `find()` to return -1. As a result, they are all considered equal. So the last one added wins, and it will be sorted before all others

I once again recommend you rethink your chosen datastructure.

(I also recommend writing some tests - then you'd be able to detect these kind of bugs much sooner, instead of relying on me to spot them)

ArminZahedani: Hmmm. I am not really in the position to make a good argument for maps here now. Using `  typedef std::vector<std::pair<std::string, double>> PipetteColor` instead of the current `map` will get rid of the ordering issue and the changes required for this are fairly minimal (from what I have seen it's mainly the `sumPipetteColors()`) . The ordering will then be the same order as the vector (or could be sorted again). This would solve the problem i believe

kees-jan: A vector of pairs would be much better, indeed. And as you say, probably only the sum and divide functions will be affected. Well, and of course the method in pipette plugin that shows the results

PR #61, opened 2020-06-03
Suppress console for Windows build

JoostvanPinxten:
Do not show a new console when the executable is started outside of a console. The stdout and stderr are still visible when the user starts scroom.exe from the command line.

This is sufficient for most users, and for extra debugging we can ask the users encountering a problem to open scroom.exe from the command line.

### DIFF
--- a/configure.ac
+++ b/configure.ac
@@ -68,7 +68,13 @@ SCROOM_ADD_CXXFLAGS_IF_SUPPORTED([-Wnon-virtual-dtor])
 SCROOM_ADD_CXXFLAGS_IF_SUPPORTED([-Woverloaded-virtual])
 SCROOM_ADD_CXXFLAGS_IF_SUPPORTED([-Wsign-promo])
 SCROOM_ADD_CXXFLAGS_IF_SUPPORTED([-Wshadow])
+
+dnl Compile with -mwindows to suppress a new console from opening if the executable is not started from a console
+dnl This does not suppress the stdout and stderr streams. These are still visible if scroom.exe is started from a 
+dnl command prompt/console.
 dnl SCROOM_ADD_CXXFLAGS_IF_SUPPORTED([-mwindows])
+AM_COND_IF([WINDOWS], 
+    [LDFLAGS="$LDFLAGS -mwindows"])
 SCROOM_TEST_ISYSTEM
 
 # Disabled warnings


JoostvanPinxten:
@kees-jan Ping! I guess you still don't get notifications about pull requests?

kees-jan:
Hi @JoostvanPinxten,
Thansk for your pull request. I made a similar remark to the students, earlier, and then they said they wanted to keep the console window for now, since it aids them in development. So when I saw your pull request, I sent them an e-mail (with you on cc) how they feel about it now.

They claim (haven't checked yet) that with this option turned on, logging also doesn't appear when starting scroom from the commandline. If that's true, then we are eliminating the only error reporting mechanism scroom currently has. Since you said before that users are likely to load files with errors, that might not be the best idea.

Hence, the current thinking seems to be to postpone merging your pull-request until after error reporting is improved.

JoostvanPinxten:
Okay, the real situation around Windows GUI with console is a bit... complicated.

I stand corrected on opening the app from the command prompt regarding showing the stdout and stderr streams. Either it is considered a [console app, or a gui app](https://docs.microsoft.com/en-us/windows/win32/api/winnt/ns-winnt-image_optional_header32). In the case of a console the command prompt waits (or spawns a new command prompt), and in the case of a gui, the command prompt returns immediately.

1) When called from a Windows Command Prompt, scroom.exe detaches and gives control back to the console immediately. This also does not allow the Command Prompt to show the output streams...
2) When scroom.exe is piped to... say... cat `scroom.exe | cat` then the output is shown when scroom exits and gives control back to the console.
3) The same as 2) happens when you redirect the output stream to file `scroom.exe > stdout.txt 2> stderror.txt`. It only flushes at the end, and does not update realtime...
4) When called from an *MSYS2 command prompt* the output streams ARE shown almost immediately.

So you are correct that this may not be the best solution... Yet strangely enough for development practices it may be sufficient to start the app from the MSYS2 command prompt.

RoanH:
Since we were asked to give our opinion on this I'll do so now.

I think that it is important to have access to errors on Windows in a way that's accessible to users that are not very technical. I agree that showing the console at all times is not the best way to do this. From my experience users are generally not going to put a lot of time into getting debug information when submitting a crash report. Only about 1% of my users (who also submit error reports in the first place) even put in the effort to run something from the command line.

Therefore I would say that `cmd` and `PowerShell` are the only tools you can expect a user submitting a crash report to have access to. For developers (which includes us) working on scroom using MSYS2 is a perfectly fine alternative to see the log.

For a regular user of Scroom option 4 isn't really an option. Option 3 appears to work in `cmd` for me but not in `PowerShell`. Option 2 doesn't work with `cat` in `PowerShell` but does with `echo`, so you could ask a user to execute `scroom.exe | echo` which would probably be the best way to ask a user for debug information after this is merged (still only flushes after closing Scroom). Piping the output like that in `cmd` doesn't appear to work however.

Given that the workaround with `echo` appears to work and assuming that @kees-jan is fine with using this workaround to get debug information from his non technical users I would say that it is fine to merge this. 

For exceptionally long errors I should note that it is possible to redirect echo so you could ask users to execute `scroom.exe | echo > output.txt` (in `PowerShell`).

Issue #66, opened 2020-06-15
Measure tool does not follow non-square pixels


KadirMarangoz:
When measuring distance in the attached TIFF file, the measurement nicely follows the horizontal pixel resolution
but the vertical pixel resolution has a mismatch (measurement tool steps by 2/3 pixel).
Note that also the scale on the left side of the image shows incorrect pixels.
Probably the latter is related and may even be the cause of the issue.

For source and attachment see: https://gitlab.com/sep-group-2/scroom/-/issues/118

PR #67, opened 2020-06-18
Pipette Plugin

RoanH:
Hello again @keesjan,

This pull request contains the pipette plugin itself. I saw that while you merged Armin's pull request you still expect him to address some issues. Do you want him to add his commits to this pull request or make a new pull request?

As for the actual content of this pull request I'll try to highlight the changes and concerns again.

### Changes
- Added the pipette tool.
  - Registers a `SelectionListener` and `PostRenderer` as well as a new tool button.
  - Heavy computations are offloaded to a separate thread to avoid freezing the GUI.
  - For now the decision was made to only allow one pipette job to compute at the same time. It is currently not possible for the user to change the selected area when the 'Computing color values...' message is shown in the status bar.
  - The output currently reports the top-left coordinate, bottom-right coordinate, height and width of the selection followed by the color channels that were returned by the `getPixelAverages` function (precision of 2 decimal places). This is currently in line with our URD.
- Remove `setPanning` and `unsetPanning` functions from the `ChildView` class. Apparently I missed these when removing them on my last PR.
  
### Concerns
- If you would like to see any major changes to the output format we might have to discuss this with CPP. The relevant URF for this is:
URF
- Currently we are resizing the selected area to be within the image bounds. This is mostly a relic from when `getPixelAverages` did not handle selections larger than the image. The main effect of this is that the reported width/height/coordinates are currently also changed and don't reflect the visual selection (but rather the portion of the selection that is within image bounds). It would be easy to change this, but I'm not sure whether to consider it a feature or a bug.
- Similar to how the status bar is not cleared when the active tool is changed no redraw event is fired either. This causes anything drawn by a post renderer to remain on screen after switching the active tool until a redraw is triggered somehow. Similar to last time I'm, not sure whether to leave this in or have the view clean it up by triggering a redraw after switching the tool.
- We might want to update the 'Error when requesting the image data' message to something like 'Image data not available for this file type'.
- As more of an extension on the last PR, we might want to pass a `ViewInterface::Ptr` in the `onEnable` and `onDisable` methods. This way the plugin could remove/add widgets and such.

Best regards,
Roan


kees-jan:
Hi Roan,

See picture below:
image

I selected a blue pixel. It is reported as yellow. Plus it is not drawn correctly. That last thing, Canon may be able to forgive you, but the color being wrong will be a tougher discussion :wink: 

I reported this problem earlier to Sverre. I hoped you'd have fixed it by now.

RoanH:
This does indeed appear to be related to the issue Peter reported last week and which we filed an issue for #66. It does appear like we might really have to look into this afterall.

Like stated before this appears to be a more general issue and not related to just the drawing of the outline (left ruler in this case doesn't report the correct information).
afbeelding

I don't mind trying to look into this, but I'm completely unfamiliar with the the part of the code base that handles this so if you have any ideas at all or if you have time for a meeting later today that would be much appreciated.

For now I would like you to review this PR based on an image without non-square pixels as the fix for this specific issue likely won't be touching any of the files touched by this PR (except for maybe some more information being passed via the post renderer).

RoanH:
Forcefully disabling support for non-square pixels shows that at least the computation of the average pixel values is correct.

afbeelding

kees-jan:
My solution would be to not fix the ruler/redrawing issue, and try to convince Canon to live with that. Then, in `TransformPresentation::getPixelAverages()`, I'd correct the given rectangle for the aspect ratio, before passing it on to the underlying presentation. You'd have to agree with Canon on how rounding is done (I'd propose to have each pixel that is at least partially covered count completely).

Only alternative would be to disable non-square-pixel-support for tiffs that report a dpi value (and only support it for those that specify a resolution of type "None"

RoanH:
Alright...

I think that I do want to at least make an attempt fixing the issue (though I can't really afford to spend more than today on it). If I cannot make any progress I'll go with the TransformPresentation solution (though we'll probably want to discuss that with Canon).

plugins/pipette/src/pipette.cc
```diff
@@ -0,0 +1,223 @@
...
+#include <cmath>
+
+#include <scroom/pipetteviewinterface.hh>
+#include <scroom/unused.hh>
+
+////////////////////////////////////////////////////////////////////////
+// Pipette
+////////////////////////////////////////////////////////////////////////
+
+Pipette::Pipette()
```
kees-jan, line 13: Not a big fan of empty constructors/destructors. Can we delete them?

plugins/pipette/src/pipette.cc
```diff
@@ -0,0 +1,223 @@
...
+  return Scroom::Bookkeeping::Token();
+}
+
+////////////////////////////////////////////////////////////////////////
+// PipetteHandler
+////////////////////////////////////////////////////////////////////////
+
+PipetteHandler::PipetteHandler()
+{
+  selection = nullptr;
```
kees-jan, line 65: You should prefer member initializer lists for initializing member variables

plugins/pipette/src/pipette.cc
```diff
@@ -0,0 +1,223 @@
...
+
+PipetteHandler::Ptr PipetteHandler::create()
+{
+  return Ptr(new PipetteHandler());
+}
+
+void PipetteHandler::computeValues(ViewInterface::Ptr view)
+{
+  gdk_threads_enter();
+  view->setStatusMessage("Computing colors values...");
```
kees-jan, line 81: I think "color" should be singular here.

plugins/pipette/src/pipette.cc
```diff
@@ -0,0 +1,223 @@
...
+  gdk_threads_leave();
+
+  // Get the image rectangle
+  auto presentation = boost::static_pointer_cast<PresentationInterface>(view->getCurrentPresentation());
```
kees-jan, line 85: That cast should not be necessary. The returned value already is an `PresentationInterface` pointer

plugins/pipette/src/pipette.cc
```diff
@@ -0,0 +1,223 @@
...
+
+  // Get the average color within the rectangle
+  auto pipette = boost::dynamic_pointer_cast<PipetteViewInterface>(presentation);
+  if(pipette == nullptr)
+  {
+    printf("PANIC: Presentation does not implement PipetteViewInterface!\n");
+    gdk_threads_enter();
+    view->setStatusMessage("Error when requesting the image data.");
+    gdk_threads_leave();
+    enabled = true;
```
kees-jan, line 113: Why modify `enabled` here?

And... Wasn't `enabled` already `true` ?

plugins/pipette/src/pipette.cc
```diff
@@ -0,0 +1,223 @@
...
+    return;
+  }
+  auto image = presentation->getRect().toIntRectangle();
+
+  // Get the selection rectangle
+  int sel_x = std::min(selection->start.x, selection->end.x);
+  int sel_y = std::min(selection->start.y, selection->end.y);
+  auto sel_rect = Scroom::Utils::Rectangle<int>(sel_x, sel_y, selection->width(), selection->height());
```
kees-jan, line 100: `Rectangle` does handle (and correct) negative width and heigth, so this can be simplified.

Perhaps `Rectangle` should have a new constructor that takes two `Point`s?

plugins/pipette/src/pipette.cc
```diff
@@ -0,0 +1,223 @@
...
+
+  // Intersect both rectangles to get the part of the selection that overlaps the image
+  auto rect = sel_rect.intersection(image);
+
+  // Get the average color within the rectangle
+  auto pipette = boost::dynamic_pointer_cast<PipetteViewInterface>(presentation);
+  if(pipette == nullptr)
```
kees-jan, line 107: Didn't get to that part yet, but if you checked the property (and you should) before adding the toolbar button, then this should always succeed (and should be turned into a `require()`.

plugins/pipette/src/pipette.cc
```diff
@@ -0,0 +1,223 @@
...
+    printf("PANIC: Presentation does not implement PipetteViewInterface!\n");
+    gdk_threads_enter();
+    view->setStatusMessage("Error when requesting the image data.");
+    gdk_threads_leave();
+    enabled = true;
+    return;
+  }
+  auto colors = pipette->getPixelAverages(rect);
+
+  // If selection became null the plugin was switched off so ignore the result
```
kees-jan, line 118: If there are several threads in this class, then you should have used mutexes. And then `selection` would not be `null` unexpectedly

plugins/pipette/src/pipette.cc
```diff
@@ -0,0 +1,223 @@
...
+  enabled = true;
```
kees-jan, line 119: In the comment, you suggest that the plugin was switched of, so you re-enable it?

plugins/pipette/src/pipette.cc
```diff
@@ -0,0 +1,223 @@
...
+
+void PipetteHandler::onSelectionEnd(Selection::Ptr s, ViewInterface::Ptr view)
+{
+  if(enabled)
+  {
+    selection = s;
+
+    // Prevent more than one job
+    enabled = false;
+    Sequentially()->schedule(boost::bind(&PipetteHandler::computeValues, shared_from_this<PipetteHandler>(), view), pendingJobs);
```
kees-jan, line 171: You should definitely improve the handling of multiple threads

plugins/pipette/src/pipette.cc
```diff
@@ -0,0 +1,223 @@
...
+  }
+}
+
+////////////////////////////////////////////////////////////////////////
+// ToolStateListener
+////////////////////////////////////////////////////////////////////////
+
+void PipetteHandler::onDisable(){
+  selection = nullptr;
+  enabled = false;
```
kees-jan, line 218: .. So enabled has two functions. It deals with whether the tool is enabled, and it has a (more obscure) role in managing concurrency. Right?

plugins/pipette/src/pipette.cc
```diff
@@ -0,0 +1,223 @@
...
+PipetteHandler::~PipetteHandler()
+{
+}
+
+PipetteHandler::Ptr PipetteHandler::create()
+{
+  return Ptr(new PipetteHandler());
+}
+
+void PipetteHandler::computeValues(ViewInterface::Ptr view)
```
kees-jan, line 78: I was going to say that this file was fairly long, but this method occupies one third of it. Perhaps you should do some splitting up.

plugins/pipette/src/pipette.cc
RoanH: Sure

plugins/pipette/src/pipette.cc
RoanH: Ah, yes, I should've noticed that.

plugins/pipette/src/pipette.cc
RoanH: Yes, you're correct

plugins/pipette/src/pipette.cc
RoanH: True, this is still a relic from when it wasn't, changed.

plugins/pipette/src/pipette.cc
RoanH: It will, but not for the first view. The first window opens with an initially empty view so we cannot get the property since there is no presentation yet. At least this issue exists when using a `ViewObserver`, perhaps a different solution exists.

plugins/pipette/src/pipette.cc
RoanH: The selection is only `null` at this point if the user switched to a different tool while the background task was in progress. Currently the intent is to only have a single worker thread and the UI thread.

plugins/pipette/src/pipette.cc
RoanH: It seems like I moved this statement by accident somehow... It should be after the if statement. The plugin is switched off while pixel values are being computed since we do not want to have multiple jobs running at the same time. This statement is supposed to reenable the plugin after the background job is done, but should be after the if statement.

plugins/pipette/src/pipette.cc
RoanH: The `enable` flag toggles the handling of incoming events, so it might require a rename. But yes we want to ignore incoming events in two cases. When the plugin is disabled and while a background computation is in progress.

plugins/pipette/src/pipette.cc
RoanH: True, I moved the display logic to a different function. Depending on your response to concern 2 we might want to extract the logic to get the selected area as well.

plugins/pipette/src/pipette.cc
RoanH: Enabled is used to ignore incoming events, given that we only want to handle one job at the same time `enabled` is set to `false` while a background job is running. If the if statement here fails then we want to make sure that we don't lock up the plugin before returning.

plugins/pipette/src/pipette.cc
RoanH: This might be something to look into then, but before I change anything here I would like your opinion on concern 2.

plugins/pipette/src/pipette.cc
kees-jan: `ColormapPlugin`, for example, registers a `PresentationObserver`, and then registers a `ViewObserver` with that, such that the resulting views always already have a presentation.

plugins/pipette/src/pipette.cc
kees-jan: You still should use mutexes :stuck_out_tongue:

plugins/pipette/src/pipette.cc
kees-jan: I have a debug build, were computing pixel values over an entire image takes 14 seconds. If I were to test on my netbook, of which I spoke earlier, it would probably be much more.

So if I selected the entire image, switched to measure, switched back to pipette, and then selected another image, I would have multiple background threads running. (well, except that the `Sequentially()` threadpool only has one thread).

Let's say that I were to select an entire image, and then select measure plugin, and go do some measurements. By the time the background thread finishes, it will re-enable Pipette plugin too, and now I have two subscribers for selection events.

The list goes on. I'm a bit confused at how you can so boldly claim that there are no race conditions, when the consensus in the industry seems to be that multithreading is hard and should be avoided. A consensus that you guys seem to agree with, given that you want to use only one thread for this endeavor, when you could trivially speed it up by using more.

To wrap up
* Thread handling, as it currently is, is broken and should be rethought. No amount of mutexes will fix that.
* Mutexes are not only for protecting something from concurrent access (though I hope I've proven that that, too, will happen), but also keeps the compiler and the processor from reordering instructions, and adds cpu-cache-synchronisation instructions. In short, if you want to make any prediction about what your multithreaded program will do (and be correct), then you'd better use them.

plugins/pipette/src/pipette.cc
RoanH: I'll look at how the  `ColormapPlugin` handles this then.

kees-jan:
> Hello again @keesjan,
> 
> This pull request contains the pipette plugin itself. I saw that while you merged Armin's pull request you still expect him to address some issues. Do you want him to add his commits to this pull request or make a new pull request?

Personally, I prefer small, separate pull-requests. But whatever is convenient to you.

> * Currently we are resizing the selected area to be within the image bounds. This is mostly a relic from when `getPixelAverages` did not handle selections larger than the image. The main effect of this is that the reported width/height/coordinates are currently also changed and don't reflect the visual selection (but rather the portion of the selection that is within image bounds). It would be easy to change this, but I'm not sure whether to consider it a feature or a bug.

I think that's a feature. Now I can measure the size of an image without having to be too careful about my selection

> * Similar to how the status bar is not cleared when the active tool is changed no redraw event is fired either. This causes anything drawn by a post renderer to remain on screen after switching the active tool until a redraw is triggered somehow. Similar to last time I'm, not sure whether to leave this in or have the view clean it up by triggering a redraw after switching the tool.

Not a problem.

On the other hand, if your attemts at detecting that Pipette has been disabled er successful, you might as well clean up.

> * We might want to update the 'Error when requesting the image data' message to something like 'Image data not available for this file type'.

I'd prefer not providing the pipette tool in this case. One should not give a user a tool that cannot be sensibly used.

Or, in this case, only give the dimensions of the rectangle, and delete the measure plugin, since there's much overlap in functionality.

> * As more of an extension on the last PR, we might want to pass a `ViewInterface::Ptr` in the `onEnable` and `onDisable` methods. This way the plugin could remove/add widgets and such.

What widgets are you thinking of?

plugins/pipette/src/pipette.cc
RoanH: Yes you are correct, I realized this same edge case over dinner as well. Though I do not really call issues like this a race condition (even though maybe I should?) since they are almost completely deterministic. I for some reason decided that the `enabled` flag was not accessible from the UI thread anymore after becoming `false` but that is not true. 

So I'll fix this, though I do not see how a mutex would help with this. The general idea is to disable the handling of events when a background job is in progress and there should only ever be one background job. The `onSelectionEnd` method is executed on the UI thread and therefore as far as I know there will never be called by multiple threads executing this method at the same time. Therefore setting a dedicated flag that prevents `onSelectionEnd` from executing inside `onSelectionEnd` should prevent a second job from being submitted. Then setting unsetting this flag at the end of the background job makes it so a new job can be submitted, but at this point the only background job has also finished executing. While a mutex could be used to prevent a second job from being submitted as far as I know the UI thread would in this case lock up to wait for the mutex lock. Where/how should a mutex be used here? As far as I can tell a boolean that is not accessible to the UI thread should suffice here.

> I have a debug build, were computing pixel values over an entire image takes 14 seconds. If I were to test on my netbook, of which I spoke earlier, it would probably be much more.

What kind of image are you testing this on? I'm personally developing on a laptop with a CPU that was released 10 years ago and even then I can still compute the average color value of an image that is 14391 x 6232 pixels (URD has 10k x 10k) within about 3 seconds (URD has 15sec).

RoanH:
> > Hello again @keesjan,
> > This pull request contains the pipette plugin itself. I saw that while you merged Armin's pull request you still expect him to address some issues. Do you want him to add his commits to this pull request or make a new pull request?

> Personally, I prefer small, separate pull-requests. But whatever is convenient to you.

It makes no difference for us so we'll make a separate pull-request.
 
> > * Currently we are resizing the selected area to be within the image bounds. This is mostly a relic from when `getPixelAverages` did not handle selections larger than the image. The main effect of this is that the reported width/height/coordinates are currently also changed and don't reflect the visual selection (but rather the portion of the selection that is within image bounds). It would be easy to change this, but I'm not sure whether to consider it a feature or a bug.
> 
> I think that's a feature. Now I can measure the size of an image without having to be too careful about my selection

Fair enough.

> > * Similar to how the status bar is not cleared when the active tool is changed no redraw event is fired either. This causes anything drawn by a post renderer to remain on screen after switching the active tool until a redraw is triggered somehow. Similar to last time I'm, not sure whether to leave this in or have the view clean it up by triggering a redraw after switching the tool.
> 
> Not a problem.
> 
> On the other hand, if your attemts at detecting that Pipette has been disabled er successful, you might as well clean up.

Alright, if we want cleanup then we can either have the view handle it or pass a `ViewInterface::Ptr` to `onEnable/onDisable` so the plugin itself can ask the view to invalidate the content.

> > * We might want to update the 'Error when requesting the image data' message to something like 'Image data not available for this file type'.
> 
> I'd prefer not providing the pipette tool in this case. One should not give a user a tool that cannot be sensibly used.

You mentioned the `ColormapPlugin` earlier for this so I'll have a look at that one.

> Or, in this case, only give the dimensions of the rectangle, and delete the measure plugin, since there's much overlap in functionality.

I'll leave that decision up to you. I personally prefer the measure tool for actually measuring distance, but either is fine for the URD.

> > * As more of an extension on the last PR, we might want to pass a `ViewInterface::Ptr` in the `onEnable` and `onDisable` methods. This way the plugin could remove/add widgets and such.
> 
> What widgets are you thinking of?

The `ViewInterface` declares `addSideWidget`, `addToToolbar`, `removeFromToolbar` and `removeSideWidget`, the `onDisable/onEnable` methods seem like the ideal place to do this. It would also allow a plugin to invalidate the view content which would allow a plugin to clean up the final thing drawn by a post renderer before it is disabled.

plugins/pipette/src/pipette.cc
kees-jan: Two things

**Problems solved with mutices**
When you have a variable that is accessed from multiple threads (in your implementation, that's `jobRunning` and `selection`), then you need to use some synchronization primitive (i.e. mutices) to safeguard them. I always thought that was commonly understood, and I did not expect to have to argue the point. Do you really need me to elaborate?

**Problems not solved with mutices**
There's still the scenario that if I disable and enable Pipette, and then make a selection, all while the previous computeValues is still running, then my new selection will be ignored. I don't think that's ok.

**Your other questions**
I'm testing on the Onyx evaluation image provided by canon. (20704x11660 pixels). More importantly, I've turned off optimizations. Optimizations do a lot to speed up scroom.

Note though, that speed is not my point. It is correctness. If I chose to run scroom on a raspberry pi 1, it should still run correctly (albeit possibly unbearably slow, and maybe with some memory constraints).

However, by artificially slowing down scroom, I can now manually trigger bugs in your implementation that would be much harder to trigger (though still present) at "normal" speed. And turning off optimization is an easier way to accomplish that goal, than hooking up my raspberry pi :wink:

RoanH: So what you're actually referring to is a data race instead of a race condition which is fair and I've updated the implementation to handle this. Apparently data races also appear to have more serious side effects in C++ compared to other languages?

The case where a background job remains running is fair (and known), however this is not something we have enough time left for to handle.

RoanH:
Given that we are extremely close to the deadline we have had to reprioritize a few things. For this PR that means that it is extremely unlikely that we will have the time to implement the check for the `PipetteViewInterface` or handle the issue with a background job not being cancelled when the plugin is turned off. If either of these is a major blocker for you I suggest that you close this pull request.

PR #68, opened 2020-06-19
pixel-averages-update

ArminZahedani:
Hi @kees-jan ,
These are the updates that were requested from #60 . 

### Changes:
- removed `bps` as parameter in `TiffPresentation.cc` and of course the `create` function does not expect it anymore.
- removed `CommonOperations` as a class as it is not needed anymore.

### Concerns:
- I did not change the datastructure for PipetteColors yet. I am currently trying out some different implementations for `sumPixelValues()`. If we can assume that the order of the vectors of the two parameters (`lhs` and `rhs`) to be the same, then it becomes a simple O(n) algorithm (there may be a simple check to make sure that `lhs` is not empty). However, I am not sure whether it is fine to have this assumption here. Without the assumption, i think the algorithm becomes a bit harder (and probably asymptotically more complex which could perhaps be avoided?). I do not really have a preference here...

regards,
Armin

### DIFF
--- a/inc/scroom/layeroperations.hh
+++ b/inc/scroom/layeroperations.hh
@@ -30,20 +30,13 @@ public:
                     Scroom::Utils::Stuff cache);
 };
 
-class PipetteCommonOperations : public CommonOperations, public PipetteLayerOperations
-{
-public:
-  typedef boost::shared_ptr<PipetteCommonOperations> Ptr;
-
-public:
-  virtual ~PipetteCommonOperations()
-  {}
-};
-
-class PipetteCommonOperationsCMYK : public PipetteCommonOperations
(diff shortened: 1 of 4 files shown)


kees-jan:
> * I did not change the datastructure for PipetteColors yet. I am currently trying out some different implementations for `sumPixelValues()`. If we can assume that the order of the vectors of the two parameters (`lhs` and `rhs`) to be the same, then it becomes a simple O(n) algorithm (there may be a simple check to make sure that `lhs` is not empty). However, I am not sure whether it is fine to have this assumption here. Without the assumption, i think the algorithm becomes a bit harder (and probably asymptotically more complex which could perhaps be avoided?). I do not really have a preference here...

For simplicity, I'd assume that they are in the same order. That seems like a fair assumption, since both `lhs` and `rhs` will ultimately be created by the same `LayerOperations` object.

PR #71, opened 2020-06-27
Enable dropping files onto SCROOM to open them.

JoostvanPinxten:
This PR adds a callback that allows dragging and dropping files onto a SCROOM window to load them.

This is a proof-of-concept/WIP PR. The code has only been tested to work on Windows 10. It hasn't been tested on Linux.

Feedback is welcome.

### DIFF
--- a/gui/src/callbacks.cc
+++ b/gui/src/callbacks.cc
@@ -53,6 +53,21 @@ static std::list<PresentationInterface::WeakPtr> presentations;
 static FileNameMap filenames;
 static std::string currentFolder;
 
+void ShowModalDialog(const std::string &message) {
+  printf("%s\n", message.c_str());
+  if (gdk_display_get_default()) {
+    // We're not running headless, don't open the popup
+    // We don't have a pointer to the parent window, so nullptr should
+    // suffice
+    GtkWidget *dialog = gtk_message_dialog_new(
+        nullptr, GTK_DIALOG_DESTROY_WITH_PARENT, GTK_MESSAGE_WARNING,
+        GTK_BUTTONS_CLOSE, "%s", message.c_str());
+
+    gtk_dialog_run(GTK_DIALOG(dialog));
+    gtk_widget_destroy(dialog);
+  }
+}
+
 void on_scroom_hide (GtkWidget*, gpointer user_data)
 {
   View* view = static_cast<View*>(user_data);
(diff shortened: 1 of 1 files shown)


JoostvanPinxten:
It borrows some 'framework' code from the SCROOM-plugin-cpp repository. In particular, it has an inlined version of ShowWarning/[Show](https://github.com/SCROOM-plugin-cpp/SCROOM-plugin-cpp/blob/master/plugins/sep/sep-helpers.cc#L3). I think this piece should be made available in the SCROOM core, so that more UI feedback can be given.

gui/src/callbacks.cc
```diff
@@ -496,6 +537,18 @@ void create_scroom(PresentationInterface::Ptr presentation)
   g_signal_connect (static_cast<gpointer>(drawingArea), "button-release-event", G_CALLBACK (on_button_release_event), view.get());
   g_signal_connect (static_cast<gpointer>(drawingArea), "scroll-event", G_CALLBACK (on_scroll_event), view.get());
   g_signal_connect (static_cast<gpointer>(drawingArea), "motion-notify-event", G_CALLBACK (on_motion_notify_event), view.get());
+
+
+  char * uriList = strdup("text/uri-list");
```
JoostvanPinxten, line 542: What is the right way to instantiate such a GtkTargetEntry from a constant string in C++...?

JoostvanPinxten: And ping @kees-jan

gui/src/callbacks.cc
```diff
@@ -438,6 +441,44 @@ void find_or_create_scroom(PresentationInterface::Ptr presentation)
   create_scroom(presentation);
 }
 
+void
+onDragDataReceived(GtkWidget *, GdkDragContext *, int, int,
+                        GtkSelectionData *seldata, guint, guint,
+                        gpointer)
+{
+  printf("Dropping file(s) onto SCROOM %s\n", (gchar*)seldata->data);
```
kees-jan, line 449: I try to avoid using all caps. So I'd prefer `scroom`, though `Scroom` would be fine too.

gui/src/callbacks.cc
```diff
@@ -438,6 +441,44 @@ void find_or_create_scroom(PresentationInterface::Ptr presentation)
...
+	if(filename.empty() || filename.rfind("file:///", 0) == std::string::npos) continue;
+	
+	try {
+	  load(filename.substr(8));
+	} 
+	catch (std::invalid_argument & ex) 
+	{
+	  boost::format warningFormat =
+          boost::format("Warning: unable to load file %s") % filename;
+      printf("%s\n", warningFormat.str().c_str());
```
kees-jan, line 467: Indentation seems off here...

gui/src/callbacks.cc
```diff
@@ -438,6 +441,44 @@ void find_or_create_scroom(PresentationInterface::Ptr presentation)
...
+      if (gdk_display_get_default()) {
+        // We're not running headless, don't open the popup
+        // We don't have a pointer to the parent window, so nullptr should suffice
+        GtkWidget *dialog =
+          gtk_message_dialog_new(nullptr, GTK_DIALOG_DESTROY_WITH_PARENT, GTK_MESSAGE_WARNING,
+        						 GTK_BUTTONS_CLOSE, warningFormat.str().c_str());
```
kees-jan, line 473: As a rule, I'm not a big fan of modal dialogs, since they block scroom, even in the case where there might be other work to do (such as loading other files). Thus, they violate the "as fast as possible" property I'd like scroom to have.

However, I agree that error reporting needs to improve, so I guess for now they have to do.

kees-jan:
Hi Joost,

Thanks for submitting this pull request! No need to mention me any more. I straightened out my github settings a couple of weeks ago, and now I see everything :wink: 

Having said that, I'm currently working on getting the last pieces of work of the students merged. Not everything that was shown at the customer acceptance is in my repo yet, and I'm currently fixing that. Since I'm also backlogged in some other areas, it'll be some time before I can take an in-depth look at this. Don't get me wrong, when this is ready to merge, I'll happily push the button, but if you need any assistance in getting it ready, I have to beg for your patience.

Groetjes,

Kees-Jan

gui/src/callbacks.cc
kees-jan: Looks like gtk makes a copy of anything it needs. Valgrind reports this one as a leak :wink:

kees-jan:
Tested on linux (and got it to work). Processed my own review comments :wink:. I think I answered all questions. Looks like it is ready to merge. Do you agree?

gui/src/callbacks.cc
JoostvanPinxten: It wouldn't compile without that on my machine. It's a quirky interaction between C and C++ here. But indeed I forgot to free the data again...

JoostvanPinxten:
Thanks for cleaning this up! I am not familiair with the gtk/gobject functions, so I missed those utility methods and 'rolled my own'. Good find for the portable extract uris!

I agree that this could definitely be merged now. I think adding this kind of small UI things make daily usage of Scroom just that little bit more convenient!

gui/src/callbacks.cc
kees-jan: I, too, had some complaints about it not being allowed to convert from string literals to char pointers. But the compiler does seem to accept the char array I'm currently using.

kees-jan:
Can't take al the credit. My first try didn't yet use those functions and instead had a `#ifdef WIN32` I didn't like. A friend of mine pointed out this alternative.

PR #83, opened 2021-05-28
Gtk update

JelleBootsma:
GTK update without ruler


kees-jan:
For me (ubuntu 18.04) this crashes with pretty much the same symptoms you describe for the Windows build (i.e. only logging "bootstrapping scroom" and not much else.

Perhaps that is worth looking into. I usually find Linux easier to debug than windows.

[scroom-gtk3-crashlog.txt](https://github.com/kees-jan/scroom/files/6565981/scroom-gtk3-crashlog.txt)

gui/src/callbacks.cc
```diff
@@ -443,10 +443,15 @@ void on_scroom_bootstrap(const FileNameMap& newFilenames)
   }
 
 
-  aboutDialogXml = glade_xml_new(xmlFileName.c_str(), "aboutDialog", nullptr);
+  aboutDialogXml = gtk_builder_new();
+  gchar** obj    = new gchar*[1];
+  obj[0]         = "aboutDialog";
+  gtk_builder_add_objects_from_file(aboutDialogXml, xmlFileName.c_str(), obj, NULL);
```
kees-jan, line 449: Could you have a look at memory management? I think we're leaking the GtkBuilder we're creating here, as well as the `obj` array we allocated.

The `obj` array should be null-terminated, i.e. `obj[1]` should be `NULL`. `gtk_builder_add_objects_from_file()` will look at `obj[i]` for ever larger values of `i`, and try to construct those, until it encounters an `obj[i]` that equals `NULL`.

This might very well be the cause of the crash your're seeing.

My recommendation would be to make `obj` a `std::vector<gchar>`, and then `push_back()` both `"aboutDialog"` and `nullptr`, and then pass `obj.data()` to `gtk_builder_add_objects_from_file()`.

Also, I think the call to `gtk_builder_add_object_from_file()` should perhaps be postponed until after we check `aboutDialogXml` for null, further down (although the api documentation suggests that `gtk_builder_new()` will never return null)

Lastly, the documentation for `gtk_builder_new()` suggest we should be using `gtk_builder_new_from_file()` instead?

inc/scroom/gtk-helpers.hh
```diff
@@ -42,9 +42,9 @@ namespace Scroom
 
     void useRecursiveGdkLock();
 
-    inline GdkRectangle createGdkRectangle(int x, int y, int width, int height)
+    inline cairo_rectangle_int_t createGdkRectangle(int x, int y, int width, int height)
```
kees-jan, line 45: We should probably rename that method now. Perhaps `createCairoIntRectangle()` ?

out.txt
```diff
@@ -0,0 +1,7274 @@
+[  4%] Built target threadpool
```
kees-jan, line 1: I don't think this file should have been checked in ;-)

kees-jan (commented):
Looks good!
* I haven't yet been able to do any testing, due to scroom crashing on startup
* Could you please have a look at the warnings that are being generated during compilation?
* It seems that the github action build for 18.04 is failing?
* Could you have a look at the static analysis warnings that are being generated?
Thanks!

out.txt
JelleBootsma: Removed

kees-jan:
Perhaps we should rename `External/gtk2` to `External/gtk`

kees-jan:
Could you please not resolve conversations?

I'm periodically checking if there is any work remaining, and if you guys resolve conversations, then I have to check the resolved conversations as well ;-)

out.txt
JelleBootsma: Sorry about resolving.
They are now unresolved

inc/scroom/rectangle.hh
```diff
@@ -61,9 +61,9 @@ namespace Scroom
         , vertically(rect.getVertically())
       {}
 
-      [[nodiscard]] GdkRectangle toGdkRectangle() const
+      [[nodiscard]] cairo_rectangle_int_t toGdkRectangle() const
```
JelleBootsma, line 64: Should probably also be renamed

out.txt
kees-jan: No biggie. They were, after all, correctly resolved ;-)

JelleBootsma:
Closed and replaced by pull request [84](https://github.com/kees-jan/scroom/pull/84)

PR #84, opened 2021-06-09
GTK3 Update + Custom ruler

tobiasvanelk:
This PR includes the GTK3 update and a custom ruler implementation to replace the GTK2 ruler widget.

Remaining issues GTK3 (that we are aware of):
 * Opening the open file dialog crashes the application on Windows.
 
We are still looking into these issues, but we decided to make a PR to get some feedback on the new custom ruler.
I have written a new class `Ruler`, contained in the `gui/src/ruler.cc` and `gui/src/ruler.hh` files, which imitates the old GTK2 widget as close as possible.
(I have attempted to branch from `next` and include the just ruler in the GTK2 version to make the PR smaller and clearer, but unfortunately, I am using some GTK3 functions, so I was unable to so. Apologies!)

Remaining issues custom ruler:
 * Clang-Tidy warns about the Ruler::Ptr and how it could possibly cause a memory leak. I've basically copied the way the `View` class creates a point in its `create()` function, in an attempt to stay consistent with the rest of Scroom, but I may be mishandling the pointer.
 * We are using the `G_CALLBACK` macro and similar to cast between GTK objects. This is used a lot in Scroom already and seems to be recommended by the official GTK documentation. Clang, however, warns against using C-style casts. Is this safe to ignore?
 * Unit tests still need to be written.
 * (The ruler contains no marker for the cursor position. The old ruler did have a marker, but it did not seem to actually respond to mouse movements and therefore did not serve much purpose, so we have not implemented the cursor marker yet.)


kees-jan:
Hi Tobias,

Thanks for the pull-request!

> Remaining issues GTK3 (that we are aware of):
> 
> * Opening the open file dialog crashes the application on Windows.
That, and there are a truckload of new warnings in the build that I'd like to get rid off, again. Be sure to double check the review comments on that other pull request (I think there's an unused variable remaining.

> Remaining issues custom ruler:
> 
> * Clang-Tidy warns about the Ruler::Ptr and how it could possibly cause a memory leak. I've basically copied the way the `View` class creates a point in its `create()` function, in an attempt to stay consistent with the rest of Scroom, but I may be mishandling the pointer.

Haven't looked into that yet. I'll do that next.

> * We are using the `G_CALLBACK` macro and similar to cast between GTK objects. This is used a lot in Scroom already and seems to be recommended by the official GTK documentation. Clang, however, warns against using C-style casts. Is this safe to ignore?

Yes. It is one of the joys of calling C-code from C++

> * Unit tests still need to be written.

That would be much appreciated

> * (The ruler contains no marker for the cursor position. The old ruler did have a marker, but it did not seem to actually respond to mouse movements and therefore did not serve much purpose, so we have not implemented the cursor marker yet.)

That's perfect! I never saw the purpose of those anyway ;-)

I just did some quick testing. No review yet. 1st impression is that this implementation is better than the Gtk2 one I'm used to. This one still shows numbers and tick marks, while the stock gtk2 one does not.
image

For bonus points: could you draw the numbers a bit higher? They are now touching the tick marks. Maybe you need to increase ruler height a bit.

Also, the three tick-mark types are not drawn exactly on top of eachother. Because of that, the larger ticks appear fatter near the bottom. This effect seems to become more pronounced when panning fast.

Ludicrously zoomed in to show the effect:
image
In this picture, there's some anti-aliasing going on as well. that's not a problem. Just the width-increase near the bottom of the tick-marks.

All in all, I'm very enthousiastic about the way this looks. Thanks again!

gui/src/ruler.cc
```diff
@@ -0,0 +1,362 @@
+#include <cmath>
+#include <iostream>
+#include "ruler.hh"
+
+////////////////////////////////////////////////////////////////////////
+// Ruler
+
+Ruler::Ptr Ruler::create(Ruler::Orientation orientation, GtkWidget *drawingArea)
+{
+    Ruler::Ptr ruler{new Ruler(orientation)};
```
kees-jan, line 10: Not sure why `clang-tidy` complains here, nor why it doesn't complain in many of the other places where we have similar constructs. The command-line version of `clang-tidy` seems fine with this, as is the CI build.

I would ignore this one for now. It'll sort itself out in time.

gui/src/ruler.hh
```diff
@@ -0,0 +1,189 @@
+#pragma once
+
+#include <gtk/gtk.h>
+#include <boost/smart_ptr.hpp>
```
kees-jan, line 4: Everywhere else, we include `boost/shared_ptr.hpp`. Not sure if there's a difference. However, in time we'll migrate from boost shared pointers to the standard library ones, and then a naive find-and-replace will miss this `#include`

gui/src/ruler.cc
```diff
@@ -0,0 +1,362 @@
...
+        if (intervalIndex == VALID_INTERVALS.size())
+        {
+            // We tried all intervals for the current n
+            intervalIndex = 0;
+            intervalN++;
+        }
+    }
+}
+
+double Ruler::mapRange(double x, double a_lower, double a_upper, double b_lower, double b_upper)
```
kees-jan, line 116: "Mapping between ranges" can really mean anything. From the computation, this looks more like a scale operation?

Also, I think the code becomes clearer if you replace `a` and `b` by `from` and `to`

gui/src/ruler.cc
```diff
@@ -0,0 +1,362 @@
...
+    int intervalN = 0;
+
+    while (true)
+    {
+        if (VALID_INTERVALS.at(intervalIndex) != 1 || intervalN != 0)
+        {
+            majorInterval = VALID_INTERVALS.at(intervalIndex) * pow(INTERVAL_BASE, intervalN);
+
+            // We check versus the width or height depending on orientation
+            double rulerSize = NAN;
```
kees-jan, line 88: `rulerSize` can be computed before the loop, right? And if you write it as `rulerSize = (orientation==HORIZONTAL)?width:height`, then it can also be `const`

gui/src/ruler.cc
```diff
@@ -0,0 +1,362 @@
+#include <cmath>
+#include <iostream>
+#include "ruler.hh"
+
+////////////////////////////////////////////////////////////////////////
+// Ruler
```
kees-jan, line 6: There's many repeated `if`s in this file both `if(orientation)` and `if(lowerToUpper)`. You might want to look into the "Strategy" pattern, to see if you can eliminate those.

gui/src/ruler.cc
```diff
@@ -0,0 +1,362 @@
...
+
+    if (drawingArea == nullptr) { return; }
+
+    update();
+
+    // We need to manually trigger the widget to redraw
+    gtk_widget_queue_draw(drawingArea);
+}
+
+void Ruler::update()
```
kees-jan, line 63: I was expecting `update()` to somehow trigger a redraw. It seems that it doesn't. That's fine, but then maybe you want to change the name to indicate what exactly it is you are updating.

gui/src/ruler.cc
```diff
@@ -0,0 +1,362 @@
...
+
+    // Index in the ruler's VALID_INTERVALS array
+    int intervalIndex = 0;
+    // Each interval is multiplied by 10 raised to the power n
+    const int INTERVAL_BASE = 10;
+    int intervalN = 0;
+
+    while (true)
+    {
+        if (VALID_INTERVALS.at(intervalIndex) != 1 || intervalN != 0)
```
kees-jan, line 83: This seems like a premature optimization that is also confusing the heck out of me. Can we get rid of this `if()`?

gui/src/ruler.cc
```diff
@@ -0,0 +1,362 @@
...
+}
+
+gboolean Ruler::drawCallback(GtkWidget *widget, cairo_t *cr, gpointer data)
+{
+
+    auto *ruler = static_cast<Ruler *>(data);
+
+    if (ruler->drawingArea == nullptr) { return FALSE; }
+
+    double width = ruler->width;
```
kees-jan, line 132: those two can be `const`, right?

Also, further down, you're both using `ruler->width` and `ruler->height` and `width` and `height`.

Wouldn't it make sense for this (static) method to directly call a non-static `Ruler::drawCallback` and bypass the problem alltogether?

gui/src/ruler.cc
```diff
@@ -0,0 +1,362 @@
...
+    GtkStyleContext *context = gtk_widget_get_style_context(widget);
+    gtk_render_background(context, cr, 0, 0, ruler->width, ruler->height);
+
+    // Draw outline along left and right sides and along the bottom
+    gdk_cairo_set_source_rgba(cr, &(ruler->lineColor));
+
+    cairo_set_line_width(cr, Ruler::LINE_WIDTH);
+    if (ruler->orientation == HORIZONTAL)
+    {
+        cairo_move_to(cr, LINE_COORD_OFFSET, 0);
```
kees-jan, line 145: Maybe for each of these three blocks a comment to indicate what you are drawing? Bonus points if you stuff each of these blocks into a function, called `drawWhateverItIsYouAreDrawing()` ;-)

gui/src/ruler.cc
```diff
@@ -0,0 +1,362 @@
...
+    }
+    else
+    {
+        lineLength = Ruler::MAJOR_TICK_LENGTH * width;
+    }
+
+    // Draw positive side of the ruler
+    if (ruler->upperLimit > 0) // If part of the range is indeed positive
+    {
+        // Draw the range [max(0, lowerLimit), upperLimit]
```
kees-jan, line 186: Why is it that we can't draw the entire range in one go?

gui/src/ruler.cc
```diff
@@ -0,0 +1,362 @@
+#include <cmath>
+#include <iostream>
+#include "ruler.hh"
+
+////////////////////////////////////////////////////////////////////////
+// Ruler
+
```
kees-jan, line 7: Wow... This is file hard to read/understand. Don't have any concrete tips on how to fix that, though.

gui/src/ruler.hh
```diff
@@ -0,0 +1,189 @@
...
+
+/**
+ * This class draws a ruler to a GtkDrawingArea.
+ * It is intended as a replacement for the old GTK2 ruler widget and is written
+ * to mimic that widget's behavior as close as possible.
+ */
+class Ruler
```
kees-jan, line 11: **Testing tip:**
This class has two responsibilities: Computing where stuff should go, and drawing stuff. Of those, you especially want to (unit) test the computing stuff. The drawing stuff is best tested manually.

However, unittesting the computation stuff is hard, because all of that is private in this class. Hence, I recommend introducing a RuleCalculator class (or something), that contains all the computation stuff (publicly), but none of the drawing stuff. Then you can unittest that class, and it can be a private member of Ruler, such that it still doesn't expose any internals you don't want the outside world to see.

kees-jan (commented):
So far, I've looked only at the new `ruler.cc` and `ruler.hh` files. I'll look at the rest when the other Gtk3 work has merged.

Be sure to also check the CI. There's formatting problems and static analysis problems to fix ;-)

kees-jan:
When zooming in a lot, there's a really minor tick for all of the odd integers. That looks a bit... eh... odd. Not sure what you can do, though.
image

gui/src/ruler.cc
kees-jan: I would put that complexity in a function that computes the first tick mark. Something like
```
int firstTickMark(int low, int tickDistance)
{
   if(low < 0) return low/tickDistance;
   return (low+tickDistance-1)/tickDistance;
}
```
Then similarly for the last tick, and then start from there...

tobiasvanelk:
I've incorporated quite a bit of the feedback listed above, although this PR is not quite ready for merging. Apologies I couldn't make more, smaller commits to reflect that. I've been developing the ruler in a separate dummy GTK project so I could work on it on my Windows machine. Working in our development VM is quite cumbersome.

* Ruler ticks are no longer drawn over each other and lines should appear clearer.
* Added drawing area parameter to the constructor.
* Fixed `shared_ptr` include (which had the side-effect of fixing the Clang-Tidy warning about the memory leak)
* Cleaned up a lot of code according to your feedback.
* Extracted some calculation functions into a separate class `RulerCalculations` to aid in unit testing.

I still need to change:

 * Still need to add the `require()` macro to the constructor to enforce the drawingArea not be a nullptr.
 * Unit tests have mostly been written, but I still need to copy them over to this project.
 * I'd still like to look into implementing the strategy pattern and possibly change the drawing code to simply draw the ruler in one go, like you suggested.

Hopefully, I'll have the unit tests incorporated by the end of the day and the remaining issues over the weekend. Thank you very much for your feedback!

kees-jan:
Hi Tobias,
I only got around to testing. No review yet. Tick mark behavior is much better than before. Thanks for that. Two remarks left (again ludicrously zoomed in to show the point:
image
* The horizontal line looks 2 pixels wide. It used to be only one. Is it possible to reduce the width?
* The vertical ticks have some anti-aliasing going on. As a result (even at "normal" zoom), they appear to not have a consistent thickness. Not sure what you can do here. Maybe disable anti-aliasing. Maybe only drawing tick marks at integer coordinates?

Thanks again!

gui/test/ruler-tests.cc
```diff
@@ -0,0 +1,370 @@
+#define BOOST_TEST_DYN_LINK
+#include <boost/test/unit_test.hpp>
+namespace utf = boost::unit_test;
+
+#include "../src/ruler.hh"
```
kees-jan, line 5: Doing `..` in a `#include` makes it harder to change the directory structure of the archive (and it breaks encapsulation).

Prefer doing `target_include_directories(src)` in the `CMakeLists.txt`

tobiasvanelk:
> Hi Tobias,
> I only got around to testing. No review yet. Tick mark behavior is much better than before. Thanks for that. Two remarks left (again ludicrously zoomed in to show the point:
> image
> 
> * The horizontal line looks 2 pixels wide. It used to be only one. Is it possible to reduce the width?
> * The vertical ticks have some anti-aliasing going on. As a result (even at "normal" zoom), they appear to not have a consistent thickness. Not sure what you can do here. Maybe disable anti-aliasing. Maybe only drawing tick marks at integer coordinates?
> 
> Thanks again!

Yes, we can change the horizontal line to be of width 1px.

I've attempted to get the tick marks to be clean 1px lines, but haven't gotten it to work perfectly so far. Integer coordinates in cairo actually map to the space between pixels, so to draw clean 1px lines, the coordinates actually need to be offset by 0.5 (that's the `LINE_COORD_OFFSET` constant). It works better than before, but like you noted, it still leaves these anti-aliased lines at certain positions/sizes, for some reason.

I am also working on simplifying the drawing code somewhat by applying some coordinate transformations pre-drawing, instead of significantly different drawing code for the horizontal and vertical ruler.

gui/test/ruler-tests.cc
```diff
@@ -0,0 +1,370 @@
...
+BOOST_AUTO_TEST_CASE(Ruler_firstTick_lowerLimit_neg123_interval_1,
+     * utf::description("Tests that a lower limit of -123 and interval 1 gives first tick position of -123"))
+{
+    BOOST_CHECK(RulerCalculations::firstTick(-123, 1) == -123);
+}
+
+BOOST_AUTO_TEST_CASE(Ruler_firstTick_lowerLimit_neg123_interval_25,
+     * utf::description("Tests that a lower limit of -123 and interval 25 gives first tick position of -125"))
+{
+    BOOST_CHECK(RulerCalculations::firstTick(-123, 25) == -125);
```
kees-jan, line 307: first tick can be before the start of the range?

gui/test/ruler-tests.cc
```diff
@@ -0,0 +1,370 @@
...
+  BOOST_CHECK(g_signal_handler_find(drawingArea, mask, sizeAllocateID, 0, nullptr, nullptr, ruler.get()) != 0);
+}
+
+///////////////
+// Testing an all-positive range
+
+BOOST_AUTO_TEST_CASE(Ruler_intervalCalculation_0_to_10_width_1920px,
+     * utf::description("Tests the interval for a range [0, 10] for a ruler of width 1920px is 1"))
+{
+    BOOST_CHECK(RulerCalculations::calculateInterval(0, 10, 1920) == 1);
```
kees-jan, line 56: All of these tests are good, in the sense that they test the functionality of the ruler. However, for someone new to the code, they are a bit hard to read.

When getting to know new code, some people prefer to read the tests, because the tests tell them how the code should be used, what they can expect from the code. Therefore, ideally, a test should refer a requirement. For this test, possibly something like "When the ruler is large, compared to the range, we use an interval of 1". When at all possible, use the requirement as the name of the test.

All tests here test a combination of (4) numbers, without explaining why those numbers are correct. In that sense they are not contributing to the understanding of the reader, and may also be a bit hard to maintain. When the requirements change, which tests should be changed?

This can be improved by not hardcoding the numberer, but by computing them. For example, given
```
int startRange = 0;
int endRange = 10;
```
you might do `int largeRulerWidth = (endRange-startRange)*MIN_TICK_DISTANCE`, and then check that the computed interval is 1. Then, from reading the code, we get a sense of why these numbers are the way they are.

Lastly, there is a lot of duplication in these tests. The numbers that are being tested are repeated in both the test name, the test description and the test implementation. I think for most of these testcases you could omit the description without reducing understandability.

kees-jan:
> I've attempted to get the tick marks to be clean 1px lines, but haven't gotten it to work perfectly so far. Integer coordinates in cairo actually map to the space between pixels, so to draw clean 1px lines, the coordinates actually need to be offset by 0.5 (that's the `LINE_COORD_OFFSET` constant). It works better than before, but like you noted, it still leaves these anti-aliased lines at certain positions/sizes, for some reason.

You might want to take a look at `CommonOperations::initializeCairo()` in `layeroperations.cc:71` for my attempt at trying to tame cairo ;-)

gui/src/CMakeLists.txt
```diff
@@ -14,7 +14,28 @@ target_sources(
...
+          PRIVATE ruler.cc
+                  ruler.hh
+                  rulerstrategies.cc
+                  rulerstrategies.hh)
+  target_link_libraries(
+          ruler_lib
+          PRIVATE project_options project_warnings
+          PUBLIC util
+                 PkgConfig::gtk
+                 Boost::system)
```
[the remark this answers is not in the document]
kees-jan: Oops... found the `add_test` call in `gui/CMakeLists.txt`. However, the test does not get run or built, so something is still missing.

Also, since you are putting the ruler in a library, why are you not (also) using that library in the gui? And move it to its own directory?

gui/src/callbacks.cc
```diff
@@ -48,7 +48,8 @@ static const std::string SCROOM_DEV_MODE = "SCROOM_DEV_MODE";
 const std::string        REGULAR_FILES   = "Regular files";
 
 static std::string xmlFileName;
-static GladeXML*   aboutDialogXml = nullptr;
+static GtkBuilder* aboutDialogXml = nullptr;
+GError*            error          = NULL;
```
kees-jan, line 52: This variable is unused, right.

gui/src/callbacks.cc
```diff
@@ -443,10 +444,16 @@ void on_scroom_bootstrap(const FileNameMap& newFilenames)
   }
 
 
-  aboutDialogXml = glade_xml_new(xmlFileName.c_str(), "aboutDialog", nullptr);
+  aboutDialogXml = gtk_builder_new();
```
kees-jan, line 447: We're still leaking this builder, right? (actually two times, due to the duplication ;-))

kees-jan (changes requested):
* Please have a look at the (github) CI build. It is still red
* Please have a look at the compiler warnings you added

gui/test/ruler-tests.cc
tobiasvanelk: Yes, it usually starts drawing from outside the drawing area, so it can still draw the minor ticks at the start, because the ruler doesn't always start with a major tick. This seemed like the simplest solution to me, instead of adding a separate case for drawing the minor ticks at the start.

kees-jan:
> Yes, we can change the horizontal line to be of width 1px.
> 
> I've attempted to get the tick marks to be clean 1px lines, but haven't gotten it to work perfectly so far. Integer coordinates in cairo actually map to the space between pixels, so to draw clean 1px lines, the coordinates actually need to be offset by 0.5 (that's the `LINE_COORD_OFFSET` constant). It works better than before, but like you noted, it still leaves these anti-aliased lines at certain positions/sizes, for some reason.

Tick marks now look amazing. For me, the horizontal line is still 2 pixels, though

gui/src/CMakeLists.txt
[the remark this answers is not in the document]
kees-jan: Not sure what changed, but ruler tests are now executing...

kees-jan:
Quoting from an e-mail to Thomas:

> Hiervan mag je dan dus de gdk_threads_* warnings laten zitten. Hetzelfde geldt voor de twee PluginInformation warnings (want zat er voorheen ook al in). Dan blijft dus voor jullie om op te lossen:
> 
> ```
>     'gdk_cairo_create' is deprecated [-Wdeprecated-declarations]
>     'gdk_window_process_all_updates' is deprecated [-Wdeprecated-declarations]
>     'gtk_widget_override_background_color' is deprecated [-Wdeprecated-declarations]
>     'gtk_widget_override_color' is deprecated [-Wdeprecated-declarations]
>     ISO C++11 does not allow conversion from string literal to 'char *' [-Wwritable-strings]
>     private field 'orientation' is not used [-Wunused-private-field]
> ```
> 
> Vat ik dat zo goed samen?

libs/tiled-bitmap/tools/measure-framerate-callbacks.cc
```diff
@@ -39,14 +39,22 @@ static void on_hide(GtkWidget*, gpointer) { gtk_main_quit(); }
 
 static gboolean on_expose(GtkWidget* widget, GdkEventExpose*, gpointer)
 {
-  cairo_t* cr = gdk_cairo_create(gtk_widget_get_window(widget));
+  cairo_region_t* re = cairo_region_create();
```
kees-jan, line 42: We'll probably see this pattern a lot. Could we create a RAII style helper class for this?

PR #86, opened 2021-06-11
Changes for CMake compiling as a submodule

JelleBootsma:
Fixes so that Scroom can be compiled as a cmake subdirectory in other repositories

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -12,7 +12,7 @@ include(cmake/ProjectVersion.cmake)
 include(cmake/StandardProjectSettings.cmake)
 include(cmake/PreventInSourceBuilds.cmake)
 
-set(PLUGIN_INSTALL_LOCATION_RELATIVE "lib/scroom")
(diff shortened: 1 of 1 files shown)


CMakeLists.txt
```diff
@@ -134,11 +144,14 @@ add_subdirectory(gui)
...
   install(CODE "set(PLUGINS \"${PLUGINS}\")")
+  if (SCROOMBINDIR)
+    install(CODE "set(SCROOMBINDIR \"${SCROOMBINDIR}\")")
+  endif()
   install(
     CODE [[
       message("Gathering dependencies for: ${PLUGINS}")
       file(GET_RUNTIME_DEPENDENCIES
-        EXECUTABLES bin/scroom.exe
+        EXECUTABLES ${SCROOMBINDIR}bin/scroom.exe
```
JelleBootsma, line 154: Add if statement around this, to clarify that SCROOMBINDIR is an optional variable.  (I believe that non existant CMake variables are rendered as empty strings, so functionally they are same)

